PR #2, opened 2016-08-27
optimize datadog backend

achille-roussel:
@f2prateek since I think you're the only one with some context on this repo you're on review duty 💃 

I've noticed that this library was causing huge GC work on a service that depends on it so I spent a bit of time digging into the root cause.

This is the first part of the fix, the datadog serialization code was causing huge amounts of memory allocations and was on a pretty busy path.

Please take a look and let me know what you think about it.

### DIFF
--- a/datadog/backend.go
+++ b/datadog/backend.go
@@ -1,13 +1,12 @@
 package datadog
 
 import (
-	"fmt"
 	"io"
 	"net"
-	"strings"
 	"time"
 
 	"github.com/segmentio/stats"
+	"github.com/segmentio/stats/iostats"
 	"github.com/segmentio/stats/netstats"
 )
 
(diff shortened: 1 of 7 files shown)


f2prateek:
LGTM!

Not entirely sure if the iostats.Formatter abstraction is neeeded - the functions could just operate on the byte array directly? i.e. instead of `write(w, f.FormatStringFunc(m.Name, sanitizeRune))` could do `write(w, formatStringFunc(buf, m.Name, sanitizeRune))`? But it does keeps the logic contained into the formatter struct/package .

Also some of the mehtods like `FormatBool` are unused.

achille-roussel:
Agree with you, the Formatter is not mandatory but it makes reusability easier across the package easier (think the statsd backend, it could really benefit from the same optimizations, not having to manage the buffer at all is nice).

FormatBool is unused although it's pretty short and there's a test for it... I just wanted parity with what the standard strconv package exposed.

Thanks for reviewing!

PR #4, opened 2016-08-29
fix panic when using http.DefaultClient.Transport

achille-roussel:
@f2prateek

### DIFF
--- a/httpstats/transport_test.go
+++ b/httpstats/transport_test.go
@@ -12,6 +12,7 @@ import (
 
 func TestTransport(t *testing.T) {
 	for _, transport := range []http.RoundTripper{
+		nil,
 		&http.Transport{},
 		http.DefaultTransport,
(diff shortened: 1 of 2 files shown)


httpstats/transport_test.go
```diff
@@ -11,40 +11,46 @@ import (
 )
 
 func TestTransport(t *testing.T) {
-	backend := &stats.EventBackend{}
-	client := stats.NewClient(backend)
-	defer client.Close()
+	for _, transport := range []http.RoundTripper{
+		&http.Transport{},
```
f2prateek: add a case for `nil`?

achille-roussel: Technically `http.DefaultClient.Transport` is `nil` but yeah that's a good call just in case it changes.

f2prateek:
LGTM!

PR #5, opened 2016-08-30
simplify go metrics

achille-roussel:
@f2prateek 

Bunch of optimizations again, get rid of useless metrics that were causing huge numbers of memory allocations.

Basically the issue was in the Collects metric was setting tags dynamically which means the tags will be copied into a new slice.

I also got rid of the stats for memory allocation breakdown per memory bucket... the profiling tools are powerful enough that we don't need these data and they're generating huuuuuuge amount of metrics.

Let me know if you'd like to see anything changed.

### DIFF
--- a/procstats/go.go
+++ b/procstats/go.go
@@ -3,7 +3,6 @@ package procstats
 import (
 	"runtime"
 	"runtime/debug"
-	"strconv"
 	"time"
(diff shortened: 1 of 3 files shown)


f2prateek:
LGTM!

PR #6, opened 2016-09-15
optional connection tags

achille-roussel:
@segmentio/infra 
@f2prateek 

Datadog reached out to us saying we were creating too many custom metrics, this change intends to give the ability to control what tags will be set on the low-level connection metrics so we can remove local_port and remote_port.

Please take a look and let me know if something should be changed.

### DIFF
--- a/netstats/conn.go
+++ b/netstats/conn.go
@@ -9,24 +9,53 @@ import (
 	"github.com/segmentio/stats"
 )
 
-func NewConn(c net.Conn, client stats.Client, tags ...stats.Tag) net.Conn {
+type ConnTag int
+
+const (
+	TagProtocol   ConnTag = 1 << 0
+	TagLocalAddr  ConnTag = 1 << 1
+	TagLocalPort  ConnTag = 1 << 2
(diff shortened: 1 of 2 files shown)


netstats/conn.go
```diff
@@ -9,24 +9,53 @@ import (
...
 
-func NewConn(c net.Conn, client stats.Client, tags ...stats.Tag) net.Conn {
+type ConnTag int
+
+const (
+	TagProtocol   ConnTag = 1 << 0
+	TagLocalAddr  ConnTag = 1 << 1
+	TagLocalPort  ConnTag = 1 << 2
+	TagRemoteAddr ConnTag = 1 << 3
+	TagRemotePort ConnTag = 1 << 4
```
yields: Do we need `TagHTTPPath` or similar? some http route paths are dynamic, like `/projects/:id` etc..

achille-roussel: We probably will, but I'm willing to wait for them to complain before making the change :P

achille-roussel:
Thanks for reviewing!

PR #7, opened 2016-10-05
obfuscate IDs in HTTP paths

achille-roussel:
@segmentio/infra 
@f2prateek 

This change intends to hide IDs in HTTP paths, it makes aggregation more powerful and hopefully should make datadog happier... they've blocked our HTTP metrics for now because they complained we had too many unique tags.

### DIFF
--- a/httpstats/metrics.go
+++ b/httpstats/metrics.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"net"
 	"net/http"
+	"path"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


httpstats/metrics.go
```diff
@@ -259,6 +260,29 @@ func parseHeaderToken(s string) (token string, next string) {
 	return
 }
 
+func isIDByte(c byte) bool {
+	return (c >= '0' || c <= '1') || (c >= 'a' && c <= 'f') || (c >= 'A' && c <= 'F') || c == '-'
```
f2prateek: What's an id byte? Just 0-1 and a-f inclusive?

f2prateek: (or better put, what does it represent)

achille-roussel: Yes, basically this would match numeric IDs, UUID and ObjectID representations. I believe there could be some false positive if a path element is named with an a-f letters word... I don't think we need to care much about it.

achille-roussel:
@f2prateek I added tests as you suggested... and fixed a bunch of stuff, still open to suggestions on how to handle / limit false positives.

f2prateek:
LGTM!

PR #8, opened 2016-11-18
V2

achille-roussel:
@rbranson 
@yields 
@f2prateek 
@thehydroimpulse 

Full refactor of the stats package to address a couple of issues:

- aggregate metrics within the process and flush them periodically to datadog, we're basically overloading our datadog agents right now so this should reduce the amount of UDP datagrams we send by multiple orders of magnitude

- redesign the high-level API to reduce the use of interfaces, Go's escape analysis cannot go through interface boundaries and that ended up causing a lot of unavoidable memory allocations on the metric paths what are usually pretty hot

Design
=====

Metrics
-------

The high-level interface exposes four different metric types and is very similar to every metric library in the world:
- **Counter:** for monotonically incrementing values
- **Gauge:** for values that may go up and down
- **Histogram:** observe distributions of values
- **Timer:** special case of histogram that offer a higher level interface to measure durations

Engine
-------

The engine is the meat of the code where the aggregation takes place, basically every metric is identified by a unique key composed from the name and tags and the engine will aggregate values it receives, the high-level metrics use three different operations to manage their aggregated state:
- **add:** mainly used by counters, adds a value to the current state of the metric
- **set:** mainly used by gauges, sets the value of the metric
- **observe:** mainly used by histograms to record a value on the metric

The engine periodically discards metrics that haven't been updated (every 10s by default) and exposes an interface for fetching its current state.

Aggregating counters and gauges is trivial with the *add* and *set* operations but in order to keep an accurate view of the timers (which are implemented on top of histograms) the engine keeps internal track of all the events generated by a timer. Clients then can choose to either aggregate them in a meaningful way.

Clients (Datadog)
-------

Different clients can be implemented to expose the state of an engine to metrics collection systems, for now there's only one for datadog because this is what we use.

The datadog client periodically fetches the state of the engine (every 1s by default) and sends one or more UDP datagrams with the dogstatsd representation of the metrics.
We use a ~64KB memory buffer to batch as many metrics as we can in one batch, because it goes through the loopback interface there should be no packet loss unless we overflow the socket's queue, which we think we won't because of the metric aggregation.

The datadog client aggregates values that belong to a histogram metric by computing the average and weighting the record based on how many actual measures it represents, this should give the best balance between efficiency and accuracy of the measures.

Integrations
------------

The even higher level APIs exposed by the package to integrate with system metric collection or HTTP client/server metrics collection have been left untouched and simply ported to compile and take advantage of the latest changes.


achille-roussel:
@rbranson
@yields
@f2prateek
@thehydroimpulse

Up? I know it's a lot of code to review, but it took longer to write ;)

yields:
lgtm!

Wondering if it's worth adding `stats.Gauge("foo")` that automatically do the whole `g := stats.G("foo"); g.Set()`?

achille-roussel:
That's already available actually via `stats.Set`

achille-roussel:
Thanks a lot for reviewing man!

PR #10, opened 2017-01-05
add .T() sugar

yields:


yields:
grr `T` is already taken :P

```
func T(name string, tags ...Tag) Timer
    T returns a new timer that produces metrics on the default engine.
```

meh

achille-roussel:
I guess you're stuck with making this part of your application if you really want to get rid of these go vet warnings...
```go
func tag(name, value string) stats.Tag { return stats.Tag{Name: name, Value: value} } 
```
Seems easy enough to write that it may not even be worth adding to the package.
Making it a local symbol to your own package also makes using way less verbose (since there's no need for a package prefix).
```
stats.Incr("metric", tag("hello", "world"), tag("cool", "beans"))
```
Seems like the most elegant solution to me...

PR #12, opened 2017-01-06
better dogstatsd utility

achille-roussel:
@calvinfo 
@yields 
@f2prateek 

Like there's no Bonnie without a Clyde, we needed a client to test this all...
```
$ ./dogstatsd
usage: dogstatsd [command] [arguments...]

commands:
 - add
 - agent
 - help
 - set
 - time

```

### DIFF
--- a/cmd/dogstatsd/main.go
+++ b/cmd/dogstatsd/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bytes"
 	"flag"
 	"fmt"
 	"log"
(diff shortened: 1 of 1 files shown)


calvinfo:
🔥

yields:
nice! -- no longer do i need to `echo <bleh> >  /dev/udp/localhost/...`, be nice if this was installed on all our machines, i usually have to do this in order to setup monitors for errors etc.. (metrics that didn't happen yet)

PR #13, opened 2017-01-06
fix datadog histogram

achille-roussel:
@calvinfo 
@yields 

Fix histogram aggregation... it was doing everything just fine but never committed what it computed :P

Also added tests to verify the behavior.


calvinfo:
💥

PR #14, opened 2017-01-19
fix counter bug

achille-roussel:
@tysonmote 
@dominicbarnes 
@calvinfo 

Indeed there was something wrong with the implementation, basically the datadog client is computing the delta between two flushes that it makes but instead of remembering the last value of the counter it was keeping the delta, which means values in the next flush would be meaningless.

I added a test that verified the behavior, it was missed before because the tests only verified that the values were right within a single flush.

### DIFF
--- a/datadog/client.go
+++ b/datadog/client.go
@@ -205,7 +205,7 @@ func metrics(state []stats.Metric, counters map[string]counter, now time.Time) [
 			}
 
 			counters[m.Key] = counter{
-				value:   value,
+				value:   m.Value,
 				modTime: now,
 			}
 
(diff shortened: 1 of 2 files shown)


calvinfo:
👍  I'll update kafka-forwarder once this is good. Good catch

tysonmote:
Cool — this fixes [the issue I was seeing](https://gist.github.com/tysonmote/7882c525fa90e9ad39c9bed8217dc2cd). Thanks for the quick turnaround!

```
2017/01/19 14:38:33 test.INCR:1|c|#program:test,version:0
2017/01/19 14:38:34 test.INCR:1|c|#program:test,version:0
2017/01/19 14:38:35 test.INCR:1|c|#program:test,version:0
2017/01/19 14:38:36 test.INCR:1|c|#program:test,version:0
2017/01/19 14:38:37 test.INCR:1|c|#program:test,version:0
2017/01/19 14:38:38 test.INCR:1|c|#program:test,version:0
```

yields:
<3

PR #15, opened 2017-01-23
datadog: log prefix

yields:
stats are not making it to DD, the only error i see is:

```
timeout
```

I'm not sure if it's coming from `stats` or not, but i thought i'll logs just in case.

@achille-roussel

### DIFF
--- a/datadog/client.go
+++ b/datadog/client.go
@@ -150,6 +150,8 @@ mainLoop:
 }
 
 func write(w io.Writer, b1 []byte, b2 []byte, changes []Metric) {
+	writer := writeLogger{w}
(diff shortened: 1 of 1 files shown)


achille-roussel:
💯 

Dial on UDP connections doesn't really connect to anything tho, it just configures the socket to send datagrams to a specific address.

I believe there should be other errors somewhere in the package that are being logged, do you mind prefixing them as well? (I believe the `Write` call on the connection should be at least).

yields:
done, can you have a look?

achille-roussel:
beautiful 👍

PR #17, opened 2017-01-25
adjust the socket buffer size on the datadog UDP socket

achille-roussel:
I noticed an issue on OSX (but it probably was happening on Linux as well) where the stats package would log an error saying the messages were too big.

It turns out the kernel discards UDP datagrams that are bigger than the kernel buffer for the socket, so this PR tries its best to allocate the largest kernel buffer for the socket and use that value as a limit when buffering dogstatsd metrics.

### DIFF
--- a/datadog/client.go
+++ b/datadog/client.go
@@ -68,7 +68,7 @@ func NewClient(config ClientConfig) *Client {
 		config.Address = DefaultAddress
 	}
 
-	if config.BufferSize == 0 {
+	if config.BufferSize == 0 || config.BufferSize > 65507 {
 		config.BufferSize = DefaultBufferSize
 	}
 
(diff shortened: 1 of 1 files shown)


yields:
confused a bit, should we flush the buffer when it fills up instead of depending on syscalls etc..?

achille-roussel:
What the thing I'm trying to do is batch as many metrics as possible per UDP datagram. The kernel refuses to send datagrams that are larger that its send buffer, so first I'm trying to set the kernel send buffer size to the same size than the user space buffer. If that doesn't work tho (assuming the kernel's max size is higher than what the code is trying to use) then it tries smaller size until it finds one that works, and otherwise falls back to the default socket buffer size). 

We could just use a small user land buffer size, but in that case we'd be sending way more datagrams, which translate to a higher load on the datadog agent.

datadog/client.go
```diff
@@ -102,24 +104,61 @@ func (c *Client) close() {
 	<-c.join
 }
 
+func socket(address string, sizehint int) (conn net.Conn, bufsize int, err error) {
```
tysonmote: It would be great to have a comment here explaining why we need this method so that anyone coming along later understands why we're doing the extra work here.

tysonmote:
Just for my edification — how much user and system CPU do the datadog agents on our hosts typically consume?

datadog/client.go
achille-roussel: Good call

achille-roussel:
I don't know the breakdown of CPU vs User, however they're between 80% to 100% total, because the agent is written in Python it's bound by the GIL and can't go above using more than one core and we're dropping datagrams. This was the original intent of this package, having stats aggregated in process and flushes batches at regular intervals.

PR #18, opened 2017-02-02
add stats.Duration

achille-roussel:
This PR adds helper functions for reporting precomputed durations.


yields:
LGTM!

what do you think about adding sugar for:

```bash
start := time.Now()
...
stats.Duration("compute", time.Since(start))
```

So we could just do:

```bash
defer stats.Timing("compute").Stop()
```

or similar anything similar, tired of all those `time.New()` everywhere haha

achille-roussel:
I believe the syntax is already supported with:
```go
defer stats.Time("compute", time.Now()).Stop()
```
```go
func Time(name string, start time.Time, tags ...Tag) *Clock
```

yields:
💥 that's slick :)

PR #20, opened 2017-02-09
don't aggregate histograms

achille-roussel:
It seems like the histogram aggregation is producing some weird results, not sure if it comes from averaging or from setting the sample rate but I'd like to get rid of it to get a clean baseline in production that we can then compare to.

Please take a look and let me know if you have any concerns.


tysonmote:
👍

achille-roussel:
Thanks!

PR #23, opened 2017-03-28
engine: add .Clock() sugar

yields:


achille-roussel:
Do you think thinks should be named something like `StartTimer` or `StartClock`. I'm worried that it may be confusing to have methods called `Timer` and `Time` on the same type.

yields:
@achille-roussel how about this:

```golang
func download() (err error) {
  defer stats.Clock("download").Stop()
  ....
  return
}
```

achille-roussel:
Sounds good to me 👍

yields:
> Fix the example ;)

done, i love you for catching it

PR #27, opened 2017-04-17
use circleci 2.0

yields:
@achille-roussel from ~2minutes to ~15seconds

screen shot 2017-04-17 at 12 08 33 pm


f2prateek:
Nice! Is there a way to use both the build image and circle 2.0? It's nice to be able to simulate circle's behaviour locally.

yields:
> Nice! Is there a way to use both the build image and circle 2.0? It's nice to be able to simulate circle's behaviour locally.

yeah, probably, we'd just need to change the `image:` thing, i'll try this out.

yields:
> Nice! Is there a way to use both the build image and circle 2.0? It's nice to be able to simulate circle's behaviour locally.

done!

dfuentes:
I think the final solution looks similar to this.  Circle 2.0 style docker image loading, with a `step` which kicks off the standardized build process.  I am going to be working on CI standardization, but probably not for another month or two, I think it will be useful to get some run time with circle 2.0 before then so we can see how well its going to work out long term.

yields:
> think it will be useful to get some run time with circle 2.0 before then so we can see how well its going to work out long term.

Yeah, this is exactly why i've been moving OSS projects to circle 2.0, they're really good candidates to test things out.

> govendor sync
> docker-compose up
> docker-compose down

What's the benefit of having those? or even installing docker-compose on this particular project? just so that it's the same with other projects?

achille-roussel:
There's no need to have govendor or docker-compose for this specific project, I was mentioning this with the idea that we'd keep being able to use one circle.yml file for all projects.

Something that we lose on this the ability to run the circle file locally (given, the chance of failure on CI are much lower with this format). The biggest risk with not having a standards is we end up with repos that use a weird build process and make it hard for new maintainers to do their job.

I'm not against the change, just would prefer if it was done as part of a larger initiative.

yields:
> I'm not against the change, just would prefer if it was done as part of a larger initiative.

Gotcha, make sense to me! -- i'm closing this until we come up with a standard for circle 2.0

PR #29, opened 2017-04-30
add MetricType.String

achille-roussel:
Useful for debugging purposes.

### DIFF
--- a/metric.go
+++ b/metric.go
@@ -19,6 +19,20 @@ const (
 	HistogramType
 )
 
+// String satisfies the fmt.Stringer interface.
+func (t MetricType) String() string {
+	switch t {
+	case CounterType:
+		return "counter"
+	case GaugeType:
+		return "gauge"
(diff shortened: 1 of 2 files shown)


f2prateek:
Should just go generate this.

achille-roussel:
I've never used go generate but I'm happy to rework it to auto-generate the values. Do you know of any good reads on this topic?

f2prateek:
Not sure how I originally discovered it, but these docs are pretty good https://godoc.org/golang.org/x/tools/cmd/stringer.

It'll generate something like https://github.com/segmentio/event-rules-counters/blob/master/precision/p_string.go.

You can "automate" it with go generate like this https://github.com/segmentio/event-rules-counters/blob/master/precision/precision.go#L10.

behaviour wise, it'll be slightly different from your current implementation. e.g. `CounterType` will print `CounterType` not `counter` as currently implemented.

PR #30, opened 2017-05-20
engine: add pkg level WithTags

yields:


f2prateek:
Feels like you could just do `stats.DefaultEngine.WithTags(tags...)` instead of `stats.WithTags(tags...)`. You save typing out `.DefaultEngine`, but probably worth it if this is a common use case.

Code LGTM!

yields:
> stats.DefaultEngine.WithTags(tags...)

Exactly what I want to avoid :P

PR #33, opened 2017-06-05
Prometheus

achille-roussel:
Hey guys, here's the implementation of a metrics handler to support feeding stats into Prometheus.

The code is not complete but it's functioning well enough and that I want to open it for review and discussions on some design decisions.

See more comments inline.

### DIFF
--- a/handler.go
+++ b/handler.go
@@ -29,3 +29,27 @@ type Flusher interface {
 	// or buffered internally.
 	Flush()
 }
+
+// StripTags returns a decorated version of handler that filters out the given
+// list of tag names from all handled metrics.
+func StripTags(handler Handler, tags ...string) Handler {
+	index := make(map[string]struct{}, len(tags))
+
+	for _, tag := range tags {
+		index[tag] = struct{}{}
+	}
+
+	return HandlerFunc(func(m *Metric) {
+		i := 0
+
+		for _, tag := range m.Tags {
+			if _, strip := index[tag.Name]; !strip {
+				m.Tags[i] = tag
+				i++
+			}
+		}
+
+		m.Tags = m.Tags[:i]
+		handler.HandleMetric(m)
+	})
+}
--- a/prometheus/append_test.go
+++ b/prometheus/append_test.go
@@ -37,12 +37,13 @@ hello_world 42
 	{
 		scenario: "simple histogram metric",
 		metric: metric{
-			mtype: histogram,
-			name:  "hello_world",
-			value: 42,
+			mtype:  histogram,
+			name:   "hello_world_bucket",
+			value:  42,
+			labels: labels{{"le", "0.5"}},
 		},
 		string: `# TYPE hello_world histogram
-hello_world 42
+hello_world_bucket{le="0.5"} 42
 `,
 	},
 
--- a/prometheus/append_test.go
+++ b/prometheus/append_test.go
@@ -51,15 +51,16 @@ hello_world_bucket{le="0.5"} 42
 		scenario: "counter metric with help, floating point value, labels, and timestamp",
 		metric: metric{
 			mtype:  counter,
+			scope:  "global",
 			name:   "hello_world",
 			help:   "This is a great metric!\n",
 			value:  0.5,
 			labels: []label{{"question", "\"???\"\n"}, {"answer", "42"}},
 			time:   time.Date(2017, 6, 4, 22, 12, 0, 0, time.UTC),
 		},
-		string: `# HELP hello_world This is a great metric!\n
-# TYPE hello_world counter
-hello_world{question="\"???\"\n",answer="42"} 0.5 1496614320000
+		string: `# HELP global_hello_world This is a great metric!\n
+# TYPE global_hello_world counter
+global_hello_world{question="\"???\"\n",answer="42"} 0.5 1496614320000
 `,
 	},
 }
--- a/prometheus/append_test.go
+++ b/prometheus/append_test.go
@@ -1,0 +1,91 @@
+package prometheus
+
+import (
+	"testing"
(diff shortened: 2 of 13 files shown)


handler.go
```diff
@@ -29,3 +29,27 @@ type Flusher interface {
 	// or buffered internally.
 	Flush()
 }
+
+// StripTags returns a decorated version of handler that filters out the given
+// list of tag names from all handled metrics.
+func StripTags(handler Handler, tags ...string) Handler {
```
achille-roussel: I've added this so we can easily filter out custom tags that go to datadog.

prometheus/handler.go
```diff
@@ -0,0 +1,136 @@
...
+// and adds it to the muxer used by the application under the /metrics path.
+type Handler struct {
+	// A map of metric name to histograms buckets used by the handler to record
+	// observed values.
+	//
+	// The buckets are of list of upper limits used to group the observed
+	// values.
+	//
+	// If a bucket is missing for a metric name the observed value is ignored.
+	HistgramBuckets map[string][]float64
```
achille-roussel: This is what I'm most unsure about, the way prometheus handles histograms is a bit different than datadog, basically you specify a set of buckets for your values and the library has to maintain counters for each buckets.

It seems like the datadog histograms are closer to the concept of summaries in prometheus (using percentiles), but the prometheus histograms also seem to allow more powerful support for aggregations.

Another point that I'm unsure about is on the design of having the mapping in the handler. I think it makes sense because it's a prometheus-specific concept but I also feel weird about having it decoupled from the place where metrics are being produced. I'm also worried about having to introduce potentially breaking changes if we need to expose the concept of histogram buckets in the top-level API.

Please share your thoughts, the last thing I want is making breaking changes, but making software that is intuitive to use is also very important.

achille-roussel: (also I just noticed the typo _Histgram_ => _Histogram_, I'll fix later otherwise github will drop the comment)

ivolo: its the job of the handler to map to the best representation in the target system, and it sounds like that is summaries, so I'd prefer this method.

ivolo:
Looks good to me dude! Let's 🚢

prometheus/handler.go
achille-roussel: That was what I thought initially as well but after reading more of the documentation on histograms in prometheus it seemed like they are much more flexible than summaries => https://prometheus.io/docs/practices/histograms/#quantiles

Basically the trade off is to be able to ask more questions about the data set but have the query computation be more expansive. Summaries are kind of "prepared queries" that answer only one question about the data set, and all the aggregation work is moved to the client side so the queries are basically free to run for the server.

The downside of histogram also appears to be addressable via the use of [recording rules](https://prometheus.io/docs/querying/rules/#recording-rules) so it appeared to be a much more powerful tool to implement.

We also have to be mindful of not limiting ourselves to the datadog model because we're familiar with it, this is a different tool and we should embrace the difference to get as much leverage as we can.

All that being said I'm open to revisiting the implementation if we realized it didn't fit what we're trying to measure.

achille-roussel:
Alright I made a couple of optimizations to the code so there's no dynamic memory allocations, I'm pretty satisfied with the state of the code now so unless someone has a concern I'll go ahead and merge.
```
$ go test -v -run _ ./prometheus/ -bench . -benchmem -memprofile /tmp/mem
BenchmarkAppendMetric/simple_counter_metric-4         	10000000	       146 ns/op	       0 B/op	       0 allocs/op
BenchmarkAppendMetric/simple_gauge_metric-4           	10000000	       147 ns/op	       0 B/op	       0 allocs/op
BenchmarkAppendMetric/simple_histogram_metric-4       	 5000000	       230 ns/op	       0 B/op	       0 allocs/op
BenchmarkAppendMetric/counter_metric_with_help,_floating_point_value,_labels,_and_timestamp-4         	 3000000	       545 ns/op	       0 B/op	       0 allocs/op
BenchmarkHandleMetric/counter-4                                                                       	 5000000	       260 ns/op	       0 B/op	       0 allocs/op
BenchmarkHandleMetric/gauge-4                                                                         	 5000000	       256 ns/op	       0 B/op	       0 allocs/op
BenchmarkHandleMetric/histogram-4                                                                     	 5000000	       250 ns/op	       0 B/op	       0 allocs/op
BenchmarkHash/standard_hash_function-4                                                                	10000000	       107 ns/op	 333.80 MB/s	      56 B/op	       2 allocs/op
BenchmarkHash/hash_function-4                                                                         	20000000	        73.0 ns/op	 493.42 MB/s	       0 B/op	       0 allocs/op
PASS
ok  	github.com/segmentio/stats/prometheus	14.271s
```

prometheus/append_test.go
```diff
@@ -0,0 +1,93 @@
...
+		metric: metric{
+			mtype:  counter,
+			scope:  "global",
+			name:   "hello_world",
+			help:   "This is a great metric!\n",
+			value:  0.5,
+			labels: []label{{"question", "\"???\"\n"}, {"answer", "42"}},
+			time:   time.Date(2017, 6, 4, 22, 12, 0, 0, time.UTC),
+		},
+		string: `# HELP global_hello_world This is a great metric!\n
```
yields: the protocol can have comments?

achille-roussel: Yes, and the `# TYPE` and `# HELP` comments are interpreted by prometheus to get the type and help message of a metric.

achille-roussel:
Thanks for the reviews guys!

prometheus/append_test.go
yields: that is pretty cool? -- can't decide if cool, or uncool haha

PR #35, opened 2017-06-12
set default buckets for histograms

achille-roussel:
This PR adds default buckets for the histograms generated by the stats sub-packages. The intent is to have things work out of the box with prometheus.

### DIFF
--- a/engine.go
+++ b/engine.go
@@ -72,7 +72,7 @@ func (eng *Engine) Register(handler Handler) {
 }
 
 // HistogramBuckets returns a map of metric names to buckets used to distribute
-// hitogram values.
+// histogram values.
 //
 // The buckets are of list of upper limits used to group the observed values.
 func (eng *Engine) HistogramBuckets() map[string][]float64 {
--- a/engine.go
+++ b/engine.go
@@ -3,6 +3,7 @@ package stats
 import (
 	"os"
 	"path/filepath"
+	"sort"
 	"sync"
 	"time"
 )
@@ -17,6 +18,7 @@ type Engine struct {
 	name     string
 	tags     []Tag
 	handlers []Handler
+	buckets  map[string][]float64
 	hmutex   sync.RWMutex
 }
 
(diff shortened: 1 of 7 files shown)


engine.go
```diff
@@ -60,9 +66,39 @@ func (eng *Engine) Handlers() []Handler {
...
 func (eng *Engine) Register(handler Handler) {
-	eng.hmutex.Lock()
+	eng.mutex.Lock()
 	eng.handlers = append(eng.handlers, handler)
-	eng.hmutex.Unlock()
+	eng.mutex.Unlock()
+}
+
+// HistogramBuckets returns a map of metric names to buckets used to distribute
+// hitogram values.
```
yields: `histogram`

engine.go
```diff
@@ -60,9 +66,39 @@ func (eng *Engine) Handlers() []Handler {
...
+//
+// The buckets are of list of upper limits used to group the observed values.
+func (eng *Engine) HistogramBuckets() map[string][]float64 {
+	eng.mutex.RLock()
+	buckets := make(map[string][]float64, len(eng.buckets))
```
yields: move above the lock?

engine.go
```diff
@@ -60,9 +66,39 @@ func (eng *Engine) Handlers() []Handler {
...
+	return buckets
+}
+
+// SetHistogramBuckets sets the buckets used for a histogram metric.
+//
+// Not all stats handler will respect the value distribution set with this
+// method, refer to the documentation of the handler for more details.
+func (eng *Engine) SetHistogramBuckets(name string, buckets ...float64) {
+	if !sort.Float64sAreSorted(buckets) {
+		panic("histogram buckets must be a sorted set of values")
```
yields: why not return an error?

engine.go
achille-roussel: If we do we can't set the size hint because accessing `len(eng.buckets)` requires holding the read lock.

engine.go
achille-roussel: I don't think this is an error that the program should be recovering from, the list of buckets has to be sorted otherwise it's not a list of buckets.

engine.go
yields: oh missed that :/

engine.go
yields: 👍

achille-roussel:
Thanks for the review guys!

PR #36, opened 2017-06-22
support trimming prefix on prometheus handler

achille-roussel:
The intent of this pull request is to provide a more idiomatic integration with prometheus when using defaults, by ignoring the root engine's namespace.

I didn't want to just ignore the metric namespace in the prometheus handler because I'm sure we have code that uses sub-engines to compose the full metric name
(e.g. `<service>.<category>.<metric>`).

### DIFF
--- a/prometheus/handler.go
+++ b/prometheus/handler.go
@@ -21,6 +21,20 @@ import (
 //
 // The handle ignores histograms that have no buckets set.
 type Handler struct {
+	// Setting this field will trim this prefix from metric namespaces of the
+	// metrics received by this handler.
+	//
+	// Unlike statsd-like systems, it is common for prometheus metrics to not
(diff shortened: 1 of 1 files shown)


prometheus/handler.go
```diff
@@ -71,6 +85,13 @@ func (h *Handler) timeout() time.Duration {
 
 // ServeHTTP satsifies the http.Handler interface.
 func (h *Handler) ServeHTTP(res http.ResponseWriter, req *http.Request) {
+	switch req.Method {
```
f2prateek: This is unrelated to trimming right?

achille-roussel: Yep unrelated to trimming, but related to having defaults that do the right thing.

achille-roussel:
Thanks for the review!

PR #37, opened 2017-06-22
prometheus: expose cumulative histogram buckets during collections

teh-cmc:
Hey there,

This fixes an issue regarding how Prometheus' histogram buckets are being collected.

In a nutshell, Prometheus' scraper expects for exposed [buckets to be cumulative](https://prometheus.io/docs/practices/histograms/#apdex-score).
Unfortunately, not complying with these expectations results in broken histogram & quantile computations. 😞 

Thanks!


achille-roussel:
That's something I had in mind, but didn't get a chance to verify yet. Thanks for submitting.

As you can see we're just getting ramped up with Prometheus, support was added very recently, which is the reason why it didn't get caught earlier.

I'm wondering if the right way to solve this would be to remove the `break` [here](https://github.com/segmentio/stats/blob/master/prometheus/metric.go#L378) instead of working around the fact that the counters aren't accumulating the values they should in the first place.

teh-cmc:
Thanks for the quick response.

> I'm wondering if the right way to solve this would be to remove the break here instead of working around the fact that the counters aren't accumulating the values they should in the first place.

I originally did the accumulation directly inside the `collect` call as it looked like the easiest entrypoint into the codebase.
Now that it's done though, I'm starting to think that doing it this way feels quite elegant too:
- the internal logic keeps using ordinary histograms, which feel way more natural to work & debug with imo
- accumulating the bucket counts at scrape time is basically free (i.e. the extra work is done only when absolutely needed)

In any case, this is such a small matter, I'm sure any decision you take will be good as good as another.

achille-roussel:
It's a good point, I guess if we don't do the accumulation at read time we also allow for an binary search optimization of the bucket to increment instead of the linear search there is right now.

I'm fine merging it, thanks for the contribution 👍

PR #40, opened 2017-06-24
expose procstats structures

achille-roussel:
The Go memory stats seem to give numbers that I either misinterpret or that aren't what I need in one of our projects. I'd like to access the memory stats exposed by the kernel to compare and potentially replace the use of runtime.Memstats.

Since I had already spent time implementing this is `stats/procstats` I thought it would be better to just reuse this code by exporting the internal types and functions, which is what this PR does.

Please take a look and let me know if something should be changed.

### DIFF
--- a/procstats/proc.go
+++ b/procstats/proc.go
@@ -90,69 +90,83 @@ func NewProcMetricsWith(eng *stats.Engine, pid int) *ProcMetrics {
 
 // Collect satsifies the Collector interface.
 func (p *ProcMetrics) Collect() {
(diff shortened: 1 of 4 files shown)


achille-roussel:
Yes the TODOs are because I didn't implement it for OSX/Windows, but still need the code to compile.

The procstats package actually exposes APIs to collect both the system metrics and the GC stats from the Go runtime (which are the only stats that Go gives us as far as I know).

Thanks for the review!

PR #41, opened 2017-06-25
optimize ftoa function in promtheus package

achille-roussel:
Profiling a service highlighted that about ~3% of all memory allocations in the application were taking place in the `ftoa` function. This pull request attempts to address the issue.

The trick is instead of generating a string for each float representation, we generate a single string with all float representations and take slices of it.

Please take a look and let me know what you think.

### DIFF
--- a/prometheus/metric.go
+++ b/prometheus/metric.go
@@ -412,7 +412,7 @@ func le(buckets []float64) string {
 		if i != 0 {
 			b = append(b, ':')
 		}
-		b = ftoa(b, v)
+		b = appendFloat(b, v)
 	}
 
 	return *(*string)(unsafe.Pointer(&reflect.StringHeader{
@@ -430,7 +430,7 @@ func nextLe(s string) (head string, tail string) {
 	return
 }
 
-func ftoa(b []byte, f float64) []byte {
+func appendFloat(b []byte, f float64) []byte {
 	return strconv.AppendFloat(b, f, 'g', -1, 64)
 }
 
--- a/prometheus/metric.go
+++ b/prometheus/metric.go
@@ -1,10 +1,12 @@
 package prometheus
 
 import (
+	"reflect"
 	"strconv"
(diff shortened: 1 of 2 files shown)


prometheus/metric.go
```diff
@@ -388,8 +395,43 @@ func (m metricBuckets) update(value float64) {
...
+// the values of the "le" label for each bucket of a histogram.
+//
+// The intent is to keep the number of dynamic memory allocations constant
+// instead of increasing linearly with the number of buckets.
+func le(buckets []float64) string {
+	if len(buckets) == 0 {
+		return ""
+	}
+
+	b := make([]byte, 0, 8*len(buckets))
```
maxence-charriere: not sure about this but should we also predict the space taken by ```':'```?

prometheus/metric.go
```diff
@@ -388,8 +395,43 @@ func (m metricBuckets) update(value float64) {
...
+func nextLe(s string) (head string, tail string) {
+	if i := strings.IndexByte(s, ':'); i >= 0 {
+		head, tail = s[:i], s[i+1:]
+	} else {
+		head = s
+	}
+	return
+}
+
+func ftoa(b []byte, f float64) []byte {
```
achille-roussel: `ftoa` stands for _float to ascii_ which is a common name for this function.

How should it be named?

prometheus/metric.go
achille-roussel: Yyou can think of it as 7 bytes for the numbers + 1 byte for the separator.

prometheus/metric.go
maxence-charriere: yeah but usually it returns a string with just the representation of the number. this append the representation to a already existing string.
if I use this ```go
str := ftoa([]byte("It would be nice..."), 42.42)
```
str would not be a valid representation of 42.42

maybe just a ```appendFloat```

PR #42, opened 2017-06-29
allow use of unsorted labels

achille-roussel:
We noticed performance issues on one of our applications due to the fact that the metric labels need to always be used in a consistent order, one way to achieve it is sorting them, another is to rely on the fact that the tags are always presented in the same order.

The intent of this pull request is to offer the program the option to skip the sorting step if it can guarantee that it won't generate metrics with reordered tags.


achille-roussel:
Thanks for the review!

PR #46, opened 2017-07-10
use atomic operations instead of mutexes in Counter and Gauge

achille-roussel:
This is the first of a couple of PRs I'm going to submit to attempt to make metric collection a bit faster.
While profiling production software I noticed that metric collection was taking an important part of the time spent, there were a couple of short wins like getting rid of some unused metrics or increasing the datadog client buffer size but now I'm down to having to optimize the code itself.

This PR changes the implementation of Counter and Gauge to use atomic operations instead of mutexes, which trims ~15% of time spent in the Counter and Gauge methods.
On top of being a quicker operation to execute this is also the kind of change which usually helps programs achieve higher throughput as there's no contention point anymore where goroutines may be paused.

Here are numbers from the benchmarks:
```
benchmark                   old ns/op     new ns/op     delta
BenchmarkCounter/Incr-4     64.4          54.8          -14.91%
BenchmarkCounter/Add-4      61.7          52.1          -15.56%
BenchmarkCounter/Set-4      59.3          52.7          -11.13%
BenchmarkGauge/Incr-4       64.7          55.0          -14.99%
BenchmarkGauge/Decr-4       62.3          55.2          -11.40%
BenchmarkGauge/Add-4        63.1          53.2          -15.69%
BenchmarkGauge/Set-4        62.8          52.5          -16.40%
```

Please take a look and let me know if you have any concerns.

### DIFF
--- a/atomic.go
+++ b/atomic.go
@@ -1,0 +1,41 @@
+package stats
+
+import (
+	"math"
+	"sync/atomic"
+	"unsafe"
+)
+
+type f64 uint64
+
+func (a *f64) ptr() *uint64 {
+	return (*uint64)(unsafe.Pointer(a))
+}
+
+func (a *f64) float() float64 {
+	return math.Float64frombits(atomic.LoadUint64(a.ptr()))
+}
+
+func (a *f64) add(f float64) float64 {
+	for {
+		p := a.ptr()
+		v := atomic.LoadUint64(p)
+		x := math.Float64frombits(v) + f
+
+		if atomic.CompareAndSwapUint64(p, v, math.Float64bits(x)) {
+			return x
+		}
+	}
+}
+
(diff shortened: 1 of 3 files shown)


abraithwaite:
Nice.  Although, what is the reason why we're keeping a copy of the metrics in memory in the stats package if it's just being sent to Datadog/Prometheus clients?

achille-roussel:
The Counter/Gauge/Histogram/Timer types are used as a way to construct metrics, it prevents the need to know te metric name and tags at the place where they are used, as well as expose extra operations (like Counter.Set), which require maintaining state. 

I have more optimisations in mind that will leverage those types as well.

achille-roussel:
Thanks for the reviews guys!

PR #47, opened 2017-07-11
Add redisstats package

otterley:
Add a new package called `redisstats` that provides a Redis server
handler and client RoundTripper.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Quick Start
 ### Engine
 
 A core concept of the `stats` package is the `Engine`. Every program importing
-the package gets a default engine where all metrics produced are aggregated.  
+the package gets a default engine where all metrics produced are aggregated.
 The program then has to instantiate clients that will consume from the engine
 at regular time intervals and report the state of the engine to metrics
 collection platforms.
@@ -193,6 +193,69 @@ func main() {
 
     // Wraps the default HTTP client's transport.
     http.DefaultClient.Transport = httpstats.NewTransport(http.DefaultClient.Transport)
+
+    // ...
+}
+```
+
+### Redis
+
+The [github.com/segmentio/stats/redisstats](https://godoc.org/github.com/segmentio/stats/redisstats)
+package exposes:
+
+* a decorator of
+  [`redis.RoundTripper`](https://godoc.org/github.com/segmentio/redis-go#RoundTripper)
+  which collects metrics for client requests, and
+* a decorator or
+  [`redis.ServeRedis`](https://godoc.org/github.com/segmentio/redis-go#HandlerFunc.ServeRedis)
+  which collects metrics for server requests.
+
+Here's an exmaple of how to use the decorator on the client side:
+```go
+package main
+
+import (
+    "github.com/segmentio/redis-go"
+    "github.com/segmentio/stats/redisstats"
+)
+
+func main() {
+    stats.Register(datadog.NewClient("localhost:8125"))
+    defer stats.Flush()
+
+    client := redis.Client{
+        Addr:      "127.0.0.1:6379",
+        Transport: redisstats.NewTransport(&redis.Transport{}),
+        Timeout:   0,
+    }
+
+    // ...
+}
+```
+
+And on the server side:
+
+```go
+package main
+
+import (
+    "github.com/segmentio/redis-go"
+    "github.com/segmentio/stats/redisstats"
+)
+
+func main() {
+    stats.Register(datadog.NewClient("localhost:8125"))
+    defer stats.Flush()
+
+    handler := redis.Handler{}
+
+    // Implement handler functions here
+
+    server := redis.Server{
+        Handler: redisstats.NewHandler(&handler),
+    }
+
+    server.ListenAndServe()
 
     // ...
 }
--- a/README.md
+++ b/README.md
@@ -247,9 +247,9 @@ func main() {
     stats.Register(datadog.NewClient("localhost:8125"))
     defer stats.Flush()
 
-    handler := redis.Handler{}
-
-    // Implement handler functions here
+    handler := redis.HandlerFunc(func(res redis.ResponseWriter, req *redis.Request) {
+      // Implement handler function here
+    })
 
     server := redis.Server{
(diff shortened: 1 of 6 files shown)


redisstats/handler_test.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/segmentio/redis-go"
+	"github.com/segmentio/stats"
+)
+
+func TestHandler(t *testing.T) {
+	metricHandler := &testMetricHandler{}
+	e := stats.NewEngine("")
+	e.Register(metricHandler)
+
+	wrappedRedisHandler := NewHandlerWith(e, &testRedisHandler{})
```
achille-roussel: The variable name here is quite verbose, Go guidelines usually recommend using short names as long as it doesn't impact readability. For example here we're already in the redisstats package so repeating that it's a redis handler is kind of redundant.

That being said, naming is not an exact science so feel free to leave it as it is if you feel like it helps better communicate what the code is about.

README.md
```diff
@@ -197,3 +197,66 @@ func main() {
...
+import (
+    "github.com/segmentio/redis-go"
+    "github.com/segmentio/stats/redisstats"
+)
+
+func main() {
+    stats.Register(datadog.NewClient("localhost:8125"))
+    defer stats.Flush()
+
+    handler := redis.Handler{}
```
achille-roussel: Sorry I missed that earlier, here we should replace with something like this:
```go
handler := redis.HandlerFunc(func(res redis.ResponseWriter, req *redis.Request) {
  // ...
})
```
and bellow
```go
server := redis.Server{
    Handler: redisstats.NewHandler(handler),
}
```
`redis.Handler` is an interface and therefore cannot be instantiated.

redisstats/handler.go
```diff
@@ -0,0 +1,77 @@
...
+	switch e := w.err.(type) {
+	case *net.OpError:
+		h.engine.Incr("errors",
+			stats.Tag{Name: "command", Value: req.Cmd},
+			stats.Tag{Name: "type", Value: "network"},
+			stats.Tag{Name: "operation", Value: e.Op})
+	default:
+		if e == nil {
+			h.engine.ObserveDuration("roundTripTime", endTime.Sub(startTime),
+				stats.Tag{Name: "command", Value: req.Cmd})
```
achille-roussel: Should we make the tag name consistent between `commands` and this metric? It's called `name` above and `command` here.

README.md
```diff
@@ -15,7 +15,7 @@ Quick Start
 ### Engine
 
 A core concept of the `stats` package is the `Engine`. Every program importing
-the package gets a default engine where all metrics produced are aggregated.  
+the package gets a default engine where all metrics produced are aggregated.
```
achille-roussel: Was this auto-fixed by your editor? The trailing double-space at the end of a line is a newline marker in github's markdown syntax.

otterley: Yes, my editor auto-fixes them.  What was the purpose of the newline here?

redisstats/handler.go
otterley: `commands` is a count of commands where the command name is the dimension.  It feels awkward to call the name dimension `command`, since it's practically the same name as the metric name itself.  Does that make sense?

otterley: Alternatively, if we can come up with a different metric name, that would solve your concern since we could then rename its dimension.

achille-roussel: Maybe `requests`?

redisstats/handler.go
```diff
@@ -0,0 +1,77 @@
...
+
+	switch e := w.err.(type) {
+	case *net.OpError:
+		h.engine.Incr("errors",
+			stats.Tag{Name: "command", Value: req.Cmd},
+			stats.Tag{Name: "type", Value: "network"},
+			stats.Tag{Name: "operation", Value: e.Op})
+	default:
+		if e == nil {
+			h.engine.ObserveDuration("roundTripTime", endTime.Sub(startTime),
```
otterley: Stylistically, I've never put the units in a metric name.  I don't see why we can't, but it's uncommon.  If we do this, I'd like to make it consistent across all collected metrics.  I'd like to punt on that decision for now.

redisstats/handler.go
otterley: :+1:

redisstats/handler.go
achille-roussel: The prometheus guidelines actually recommend to do this. I've found it very helpful when going through metrics of existing software, whenever there's a unit it makes it easy to understand what we're looking at without having to pull out documentation or source code.

README.md
achille-roussel: Just styling on the text.

redisstats/handler.go
otterley: :+1:

redisstats/testing.go
```diff
@@ -0,0 +1,44 @@
...
+	h.Lock()
+	c := *m
+	c.Tags = append([]stats.Tag{}, m.Tags...)
+	c.Time = time.Time{} // discard because it's unpredicatable
+	h.metrics = append(h.metrics, c)
+	h.Unlock()
+}
+
+func validateMetric(t *testing.T, metric stats.Metric, name string,
+	tags []stats.Tag, metricType stats.MetricType) {
```
otterley: :+1:

redisstats/handler_test.go
otterley: :+1:

redisstats/handler_test.go
```diff
@@ -0,0 +1,83 @@
...
+	validateMetric(t, metricHandler.metrics[7], "errors",
+		[]stats.Tag{
+			{Name: "command", Value: "SET"},
+		}, stats.CounterType)
+}
+
+type testRedisHandler struct{}
+
+func (*testRedisHandler) ServeRedis(res redis.ResponseWriter, req *redis.Request) {
+	res.Write("+OK\r\n")
```
otterley: :+1:

README.md
otterley: :+1:

otterley:
@achille-roussel updated per your feedback

PR #49, opened 2017-07-23
add influxdb client

achille-roussel:
This PR adds support for InfluxDB to the stats package. The intent here is to have a InfluxDB client so I can start experimenting with it without having to rewrite any instrumentation code.

The client API is very similar to the one for datadog, basically:
```go
stats.Register(influxdb.NewClient("localhost:8086"))
```

I spent a bit of time profiling and optimizing the code, basically I was thinking of taking two different approaches to manage the batches of metrics:
- manage a list of buffers, each containing the serialized representation of a metric
- capture the metric objects themselves in a list which is then serialized when the batch is flushed

I ended going for the first solution because it was the most time-efficient implementation (~200ns per metric), however it requires sorting the tags when handling the metrics, which is on the hot path here... if sorting ends up being a bottleneck we can revisit and see if we can optimize the second approach.
The implementation also is fully wait-free, it uses atomic operations to synchronize concurrent operations on shared resources, this prevents having a high contention point if we were using a mutex in HandleMetrics.

Please take a look and let me know if you'd like to see anything changed.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,4 +1,6 @@
 machine:
+  pre:
+    - curl -sSL https://s3.amazonaws.com/circle-downloads/install-circleci-docker.sh | bash -s -- 1.10.0
   services:
     - docker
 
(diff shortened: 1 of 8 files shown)


achille-roussel:
Thanks for the review!

PR #50, opened 2017-07-23
optinally allow http path tags in datadog client

achille-roussel:
Historically we disabled the `http_req_path` tag that the httpstats package sets on metrics because there were cases where we had paths that embedded user IDs with a cardinality higher than what datadog can handle.

However there are cases where we have more control over what the HTTP requests path looks like, which means we could leverage the tag in those cases.

This PR makes it possible to lift the restriction on the `http_req_path`.

### DIFF
--- a/datadog/client.go
+++ b/datadog/client.go
@@ -30,6 +30,16 @@ type ClientConfig struct {
 
 	// BufferSize is the size of the output buffer used by the client.
 	BufferSize int
+
+	// Datadog has complained numerous times that the request paths
+	// generate too many custom metrics on their side, setting this
+	// flag to true strips out the http_req_path tag found in metrics.
+	//
+	// Programs that know they won't produce high-cardinality tags for
+	// http_req_path can enable the tag.
+	//
+	// The default value is false.
+	EnableHttpPathTag bool
 }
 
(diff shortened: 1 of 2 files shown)


datadog/client.go
```diff
@@ -30,13 +30,25 @@ type ClientConfig struct {
...
+
+	// Datadog has complained numerous times that the request paths
+	// generate too many custom metrics on their side, setting this
+	// flag to true strips out the http_req_path tag found in metrics.
+	//
+	// Programs that know they won't produce high-cardinality tags for
+	// http_req_path can enable the tag.
+	//
+	// The default value is false.
+	EnableHttpPathTag bool
```
yields: `s/Http/HTTP/` to follow golang guidelines :)

f2prateek:
The Boolean flag thing sounds great for the simple case. I wonder if it's a good idea to accept a callback function that is provided the http path, and returns a tag ("" for no tag) or tags (nil for no tags).

datadog/client.go
achille-roussel: 👍 I'll change it

achille-roussel:
The callback option sounds good, I'll revisit this 👍

PR #52, opened 2017-08-01
V4

### LINKED ISSUE #51: Proposal for a new design
I'm finding it harder and harder to have metric collection not be the bottleneck in some high-performance applications that I'm working on. I want to open the discussion on a new design to make metric collection 10 to 100x faster while keeping high-level abstractions to easily instrument applications.

Here are the most common issues:
- Serializing the metrics is done _inline_ when metrics are produced, it doesn't matter how fast we can make the serialization routines it's inefficient because it deals with memory that has nothing to do with the main work that the application is doing, likely causes a lot of cache misses, and is a single contention point for goroutines producing the metrics.
- When no serialization is being done inline (for example with prometheus) the stats handler needs to maintain an in-memory representation of the state, this is equally inefficient as it requires hitting large amounts of memory that are out of the main working set, doing map lookups, synchronizations, and puts constraints on the ordering of tags.

Here are a couple of ideas I have on what we can do:
- Batching seems necessary, allowing multiple metrics to be submitted at once from the application to the stats client would prevent frequent back and forth between the main code path and the stats production layer.
- The application often knows when it's OK to pass a batch of metrics to be recorded, the API should leverage this. For example, measuring HTTP requests, all metrics for a request/response can be aggregated before being sent with a single call after the request is fully handled, moving the metric serialization out of the critical path.
- Requiring tags to be sorted before the metrics is produced seems like the only way to avoid the cost at the lower layer. The program can then then order tags once for all metrics of a batch.
- Batches of metrics can be efficiently represented by structs types like:
```go
type HTTPMetrics struct {
  Message struct {
    Count int64 `metric:"count" type:"counter"`
  
    Request struct {
      HeaderLength int64 `metric:"header.bytes" type:"histogram"`
      BodyLength   int64 `metric:"body.bytes" type:"histogram"`
      ...
    }

    Response struct {
      HeaderLength int64 `metric:"header.bytes"`
      BodyLength   int64 `metric:"body.bytes"`
      RTT          time.Duration `metric:"rtt.seconds"`
      ...
      Status int `tag:"http_res_status"`
    }

    // top-level tags get inherited by sub-metrics
    Path string `tag:"http_req_path"`
    ...
  } `metric:"http.message"`
}
```
There are a couple of reasons why I think this approach can be more powerful:
1. When generating the metric, the program just sets the value of each field, it's hard to make the code more efficient than a couple of assignments and increments that will likely be collocated to the actual working set of the application (metrics values can be embedded directly into structures of the program, or even allocated on the stack).
2. The implementation will need to use reflection to read the metric structs, this can be made highly efficient by doing runtime _compilation_ of the reflection code (this technique is used in `encoding/json` and `github.com/segmentio/objconv` for example). This would make it possible to generate highly efficient ways to render the sorted list of tags. I also means that we can do more optimizations to like generating efficient hash tables to store and aggregate the metrics before flushing them to the network.
3. One of the nice things is the metric structs don't depend on special types, they use basic Go types or types from the standard library, so exposing metrics from library packages becomes trivial and is fully decoupled from the stats package itself.
4. Using structs brings the more declarative design (instead of the very imperative approach we have now), it also helps the developer thinks about how things are measured as part of the data structures of the program, not as random additions in the middle of the code.

Now this is a pretty big shift from the way we've been doing instrumentation, so I have a couple of questions:
- What do you guys think about the idea?
- Should this be part of stats? A subpackage? Or a different repository? The reason I'm asking is because I'm concerned about the high complexity of maintaining the code to support both the existing API and this one. Maybe the right path is to build this into a different package, then backport it if the approach is successful, or create ways to bridge between the two.

Let me know!

achille-roussel:
Fixes #51 

This PR moves the package to v4, implementing the new API for producing batches of metrics from struct types.

I've updated the README with a couple of instructions on the breaking changes that are introduced.


README.md
```diff
@@ -9,6 +9,83 @@ Installation
...
+m := funcMetrics{}
+m.calls.count = 1
+m.calls.time = callTime
+
+// Prior to v4 this could have been:
+//
+//   stats.Incr("func.calls.count")
+//   stats.ObserveDuration("func.calls.time", callTime)
+//
+stats.Report(m)
```
f2prateek: I'm less excited by this new API. Depending on whether `funcMetrics` is reused or not, it adds 8 extra lines of code. Though in the case it is reused, it could offer more consistency.

f2prateek: To clarify a bit more, I think the struct approach is great for complex cases (like the HTTP example you originally shared), or when the metric might be reused. But I think locking the API to just this makes it hard for adoption and simpler one-off metrics.

achille-roussel: I agree, the existing methods are still available so we should pick what works the best for the use case we have. 
The intent is definitely to reuse the struct or pass it around to collect metrics of multiple operations before reporting the metrics. This example is simplistic to present tye new API.

achille-roussel: Maybe I should rephrase the comment to say `Equivalent to:` instead so it doesn't ambiguously mean that it's a replacement API but an addition instead.

achille-roussel:
Up? I'd love to move forward with this and start building more things on top of it.

buffer.go
```diff
@@ -0,0 +1,159 @@
...
+	// If left to zero, a size of 1024 bytes is used as default (this is low,
+	// you should set this value).
+	BufferSize int
+
+	// The Serializer used to write the measures.
+	//
+	// This field cannot be nil.
+	Serializer Serializer
+
+	buffer unsafe.Pointer
```
abraithwaite: Why does buffer need to be an `unsafe.Pointer` here?

buffer.go
```diff
@@ -0,0 +1,159 @@
...
+		if n < 0 || (n-l) > m {
+			// Only the goroutine that first overflows the buffer size is in
+			// charge of writing the buffer, the others go back and try to
+			// acquire a new one to copy their measures into.
+			continue
+		}
+
+		if h.compareAndSwapBuffer(b, nil) {
+			b.m.Lock() // wait for read-locks to be released
+			b.m.Unlock()
```
abraithwaite: Shouldn't this be after `writeBuffer`?  Even if we're only locking to wait for read-locks to be released, it's possible that another read-lock gets acquired before writeBuffer get's called no?

buffer.go
```diff
@@ -0,0 +1,159 @@
...
+// Flush satisfies the Flusher interface.
+func (h *Buffer) Flush() {
+	for {
+		b := h.loadBuffer()
+		if b == nil {
+			break
+		}
+		if h.compareAndSwapBuffer(b, nil) {
+			b.m.Lock()
+			b.m.Unlock()
```
abraithwaite: We should add the above comment here as well (assuming the above code is correct, this just looks wrong to an uninformed observer).

buffer.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+func (h *Buffer) loadBuffer() *buffer {
+	return (*buffer)(atomic.LoadPointer(&h.buffer))
+}
+
+func (h *Buffer) compareAndSwapBuffer(old *buffer, new *buffer) bool {
+	return atomic.CompareAndSwapPointer(&h.buffer,
+		unsafe.Pointer(old),
+		unsafe.Pointer(new),
```
abraithwaite: TIL you can call a variable new.  Is it like this with other keywords too?

datadog/client.go
```diff
@@ -1,103 +1,176 @@
...
+		return
+	}
+	defer f.Close()
+	fd := int(f.Fd())
+
+	// The kernel refuses to send UDP datagrams that are larger than the size of
+	// the size of the socket send buffer. To maximize the number of metrics
+	// sent in one batch we attempt to attempt to adjust the kernel buffer size
+	// to accept larger datagrams, or fallback to the default socket buffer size
+	// if it failed.
```
abraithwaite: Is this a bug in Go?  I'd expect the datagram to be split up and sent separately if this is the case.

I suppose since it's UDP there's no good way for them to decide where to split the message though.

abraithwaite: I see below that this was part of the previous package, so don't change it but I'm still curious.

buffer.go
achille-roussel: It has to be an unsafe pointer so we can perform an atomic compare-and-swap on the value.
I wish we could use `atomic.Value` to get a bit more type safety but it only supports atomic load and store.

buffer.go
achille-roussel: The intent of the lock is just to wait for all read locks to be released, so we know that there are no goroutines left in the critical section controlled by the read lock. Once that is done, no more goroutines will ever produce bytes to the buffer (either because they'll see a `nil` value after the compare-and-swap, or because they'll see a full buffer).

After successfully acquiring the write lock the goroutine is the solve owner of the buffer and therefore can release the lock. I agree it's not really intuitive, the intent is to prevent goroutines that would potentially have captured the buffer before it was set to `nil` to block on the flush and therefore move forward and allocate a new buffer on the next loop iteration.

buffer.go
achille-roussel: What if I modify the code like this:
```go
if h.compareAndSwapBuffer(b, nil) {
    b.waitReleaseRLock()
    ...
}
```
?

buffer.go
achille-roussel: `new` isn't a keyword, it's a builtin function like `make`, `append`, or `copy` ;)

datadog/client.go
achille-roussel: Those are limitations of both the UDP protocol and its implementation in the kernel. But worse, it also has to fit in the receiver's buffer, otherwise the datagram would get truncated on read as well, and the sender side has no way to know how big the receiver's buffer is.

We've had a couple of performance issues due to small UDP buffers, the datadog agent supported only 1KB datagrams for long, I think they raised it to 4KB or 8KB (which is still very small). This is one of the reasons we moved to using veneur, which we configure to use a 32KB read buffer.

They say UDP stands for User Datagram Protocol, but it feels more like Unreliable Datagram Protocol ;)

abraithwaite: And they don't support TCP I suspect?  That sucks.

buffer.go
abraithwaite: I think that would be a bit more clear yeah.

datadog/client.go
achille-roussel: Yep, they only support UDP.

buffer.go
abraithwaite: Gotcha.  So one could say that this buffer is all-or-nothing?

Asking because I've used either a circular buffer implementation or just didn't serialize and used a typed slice buffer.

achille-roussel: The intent here is to have an algorithm that is mostly wait-free, this gives throughput that increases linearly with the number of cores. If we were using a single buffer (circular or not) then goroutines need to wait for flush operations to complete when the buffer is full. We need to use a dynamic number of buffers for maximum throughput (this is what the sync.Pool gives us) but we also need to avoid lock contention on transition phases, like when replacing a full buffer. Using atomic compare-and-swap we can ensure that no goroutine is contending on a lock, when they do conflicting things like having two goroutines allocate a new buffer only one will succeed and the buffer will give its buffer back to the pool.

abraithwaite: Gotcha, that all makes sense.  Totally different use cases in the situations I was referring to. 😄

achille-roussel:
Thanks a lot for the review!

PR #53, opened 2017-08-08
Add RPC stats.

f2prateek:
This adds an RPC invoker implementation that records RPC stats.

It records 2 stats:

1. Counter for `jsonrpc.call`.
2. Histogram for `jsonrpc.duration`.

Each of these contains the following tags:
1. `method` - the RPC method.
2. `resource` - the resource if the method is of the form
"resource.function".
3. `function` - the function if the method is of the form
"resource.function".
4.  `result` - either "error", "success", "panic".

### DIFF
--- a/rpcstats/invoker.go
+++ b/rpcstats/invoker.go
@@ -1,0 +1,61 @@
+package rpcstats
+
+import (
+	"context"
+	"strings"
+	"time"
+
+	"github.com/segmentio/rpc"
+	"github.com/segmentio/stats"
+)
+
+type invoker struct {
+	invoker rpc.Invoker
+	eng     *stats.Engine
+}
+
+// NewInvoker wraps a RPC invoker to produce metrics for every RPC call.
+func NewInvoker(i rpc.Invoker) rpc.Invoker {
+	return NewInvokerWith(stats.DefaultEngine, i)
+}
+
+// NewInvoker wraps a RPC invoker to produce metrics on eng for every RPC call.
+func NewInvokerWith(eng *stats.Engine, i rpc.Invoker) rpc.Invoker {
+	return &invoker{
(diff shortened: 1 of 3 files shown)


f2prateek:
I don't need this ASAP, so we can take the time to fix any issues here (e.g. switching to the new API might be a good idea).

I wasn't sure for the best place for this. An argument could be made for this to be in `rpc` as well. Let me know if you strong opinions either way.

testutils/handler.go
```diff
@@ -0,0 +1,31 @@
+package testutils
```
achille-roussel: Can we rename this `statstest`? (following the model of `http/httptest`)

testutils/handler.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"sync"
+	"time"
+
+	"github.com/segmentio/stats"
+)
+
+var _ stats.Handler = (*Handler)(nil)
+
+type Handler struct {
```
achille-roussel: There are couple of sub-packages that have redeclared a type this this over and over, do you mind following up with the PR to port those to use this type?

achille-roussel:
On your question about whether it makes more sense to put it here or in the `rpc` package, I think it's fine to have it here, so we keep everything stats related in a single location.

f2prateek:
At first glance, build failures seem unrelated to the change, I'll dig in later.

testutils/handler.go
f2prateek: Will do.

testutils/handler.go
f2prateek: Yep! i'll do them one by one.

f2prateek:
Closing this as per discussion on slack.

PR #54, opened 2017-08-08
Add statstest package.

f2prateek:
The `statstest` package contains a single `stats.Handler` implementation
that records its measures for inspection. It also implements
`stats.Flusher` so tests can inspect number of times `Flush` was
invoked.

Also removes replaces custom test handlers with the one from the new
`statstest` package.


f2prateek:
(going to use this in `rpcstats`).

f2prateek:
Fixed build errors and rebased against master.

PR #58, opened 2017-08-24
tag: add stats.T shorthand func

### LINKED ISSUE #57: Allow for tags to be map[string]string
Realized we're not doing vet on a number of repos and found this:

https://gist.github.com/abraithwaite/9660621e4f8259d3d7445d1dcabc292b

I don't see a super compelling reason to have the stats.Tag to be typed so strongly over map[string]string.

Convince me otherwise ;-)

abraithwaite:
This enables us to run vet on code without the full verbosity of the go
keyed fields for structs.

Fixes #57

### DIFF
--- a/tag.go
+++ b/tag.go
@@ -12,6 +12,12 @@ type Tag struct {
 	Value string
 }
 
+// T is shorthand for `stats.Tag{Name: "blah", Value: "foo"}`  It returns
+// the tag for Name k and Value v
+func T(k, v string) Tag {
+	return Tag{Name: k, Value: v}
+}
+
 func (t Tag) String() string {
 	return t.Name + "=" + t.Value
 }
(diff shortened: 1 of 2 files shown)


tag.go
```diff
@@ -12,6 +12,10 @@ type Tag struct {
 	Value string
 }
 
+func T(k, v string) Tag {
```
achille-roussel: Can we document this function? So we don't trade a govet warning for one in golint ;)

abraithwaite: Of course! :-)

abraithwaite:
Protip for migration:
```
:%s/\vstats\.Tag\{"(\w+)", ("?[A-Za-z -\.\(\)_]+"?)\}/stats.T("\1", \2)/g
```

achille-roussel:
Thanks for the sed, that's just what I needed :D

PR #59, opened 2017-08-29
fix procstats counters

achille-roussel:
I noticed issues where some counters just keep incrementing to values that don't make sense at all... the problem is counters should report the delta since the last flush, not their incrementing value.

This PR addresses the problem.


achille-roussel:
Thanks!

PR #61, opened 2017-08-29
add clock

yields:
basically copy pasted from v3, i think in v3 `total` stamp would actually just measure from `.last`, which means it's not really "total", so changed it to measure from `.first`.

no timing assertions, but willing to add them if you folks think it's necessary.

the goal is to get this:

```
upload stamp=grayscale (2s)
upload stamp=compress (1s)
upload stamp=total (3s) (uses `.first`)
```

I _think_ v3 used to report this:

```
upload stamp=grayscale (2s)
upload stamp=compress (1s)
upload stamp=total (1s) (uses `.last`)
```

### DIFF
--- a/clock.go
+++ b/clock.go
@@ -1,0 +1,53 @@
+package stats
+
+import "time"
+
+// The Clock type can be used to report statistics on durations.
+//
+// Clocks are useful to measure the duration taken by sequential execution steps
+// and therefore aren't safe to be used concurrently by multiple goroutines.
+type Clock struct {
+	name  string
+	first time.Time
+	last  time.Time
+	tags  []Tag
+	eng   *Engine
+}
+
+// Stamp reports the time difference between now and the last time the method
+// was called (or since the clock was created).
+//
+// The metric produced by this method call will have a "stamp" tag set to name.
+func (c *Clock) Stamp(name string) {
+	c.StampAt(name, time.Now())
+}
+
+// StampAt reports the time difference between now and the last time the method
+// was called (or since the clock was created).
+//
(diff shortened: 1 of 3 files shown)


clock.go
```diff
@@ -0,0 +1,53 @@
...
+// StopAt reports the time difference between now and the time the clock was created at.
+//
+// The metric produced by this method call will have a "stamp" tag set to
+// "total".
+func (c *Clock) StopAt(now time.Time) {
+	c.observe("total", now.Sub(c.first))
+}
+
+func (c *Clock) observe(stamp string, d time.Duration) {
+	tags := append(c.tags, Tag{"stamp", stamp})
```
achille-roussel: One thing you could do to avoid a dynamic memory allocation here is pre-allocate the `c.tags` slice with an extra slot at the end.

engine.go
```diff
@@ -94,6 +94,18 @@ func (eng *Engine) Observe(name string, value interface{}, tags ...Tag) {
 	eng.measure(name, value, Histogram, tags...)
 }
 
+// Clock returns a new clock identified by name and tags.
+func (eng *Engine) Clock(name string, tags ...Tag) *Clock {
+	now := time.Now()
```
achille-roussel: We should copy the tags here, otherwise it may be non-intuitive that this slice gets captured by the returned Clock.

clock.go
yields: good call, added another slot

engine.go
yields: good call, copied! :)

PR #63, opened 2017-08-30
fix procstats

achille-roussel:
There are three main changes in the PR:
- Report percentage of CPU and Memory usage of the process
- Fix CPU usage values on linux by switching to use `getrusage`
- Add implementation process metrics collection on darwin

Those things are pretty hard to test because we can never predict how much resources a process is using (plus it's very dependent on the environment), here's a trace of the tests ran on OSX:
```
	proc_test.go:26: { cpu(counter:usage.seconds=46µs, gauge:usage.percent=73.76050285421076) [type=user] }
	proc_test.go:26: { cpu(counter:usage.seconds=16µs, gauge:usage.percent=25.655827079725483) [type=system] }
	proc_test.go:26: { memory(gauge:usage.bytes=2707456, gauge:usage.percent=0.03151893615722656) [type=resident] }
	proc_test.go:26: { memory(gauge:usage.bytes=0) [type=shared] }
	proc_test.go:26: { memory(gauge:usage.bytes=0) [type=text] }
	proc_test.go:26: { memory(gauge:usage.bytes=0) [type=data] }
	proc_test.go:26: { memory.pagefault(counter:count=0) [type=major] }
	proc_test.go:26: { memory.pagefault(counter:count=6) [type=minor] }
	proc_test.go:26: { memory(gauge:available.bytes=8589934592, gauge:total.bytes=2707456) [] }
	proc_test.go:26: { files(gauge:open.count=45, gauge:open.max=7168) [] }
	proc_test.go:26: { threads.switch(counter:count=0) [type=voluntary] }
	proc_test.go:26: { threads.switch(counter:count=0) [type=involuntary] }
	proc_test.go:26: { threads(gauge:count=5) [] }
```
The values seem to make sense, let me know if anything should be changed!

### DIFF
--- a/procstats/proc_darwin.go
+++ b/procstats/proc_darwin.go
@@ -1,27 +1,122 @@
 package procstats
 
+/*
+#include <mach/mach_init.h>
+#include <mach/mach_host.h>
+#include <mach/mach_port.h>
+#include <mach/mach_traps.h>
+#include <mach/task_info.h>
+#include <mach/task.h>
+#include <sys/proc_info.h>
+#include <sys/types.h>
+#include <sys/sysctl.h>
+
+// Defined in kern/proc_info.c
+extern int proc_pidinfo(int pid, int flavor, uint64_t arg, user_addr_t buffer, uint32_t buffersize, register_t * retval);
+
+// CGO fails to call a macro that just references a global variable...
+#ifdef mach_task_self
+#undef mach_task_self
+static mach_port_t mach_task_self() { return mach_task_self_; }
+#endif
+*/
+import "C"
+import (
+	"errors"
+	"fmt"
+	"os"
+	"syscall"
+	"time"
+	"unsafe"
+)
+
 func collectProcInfo(pid int) (info ProcInfo, err error) {
-	getclktck()
-	// TODO
+	defer func() { err = convertPanicToError(recover()) }()
+
+	if pid != os.Getpid() {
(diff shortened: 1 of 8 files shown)


procstats/proc_darwin.go
```diff
@@ -1,27 +1,122 @@
...
+#include <mach/task_info.h>
+#include <mach/task.h>
+#include <sys/proc_info.h>
+#include <sys/types.h>
+#include <sys/sysctl.h>
+
+// Defined in kern/proc_info.c
+extern int proc_pidinfo(int pid, int flavor, uint64_t arg, user_addr_t buffer, uint32_t buffersize, register_t * retval);
+
+// CGO fails to call a macro that just references a global variable...
```
abraithwaite: ouch.

abraithwaite:
Why do we care about darwin?  Is there a darwin host which _needs_ to read these?

If we don't need to support it, we should just fill the functions with mocks and state that it's not supported.

achille-roussel:
We don't _need_ it... but it's nice to have, especially when testing programs locally that depend on this package (think something that exposes its metrics to a local prometheus or similar).

Sorry I mistakenly hit merge, feel free to leave more comments and I'll address them.

abraithwaite:
No biggie.  You're right it's nice to have.  I just dislike cgo.  We can always remove it later if it's too much to bear.

achille-roussel:
Agree, CGO is only used of OSX tho, the linux version of the code doesn't depend on any C code.

PR #64, opened 2017-08-31
report total CPU usage

achille-roussel:
Reporting the total value allows min/max metrics to be actually meaningful, otherwise computing `min(user) + min(sys)` adds metrics that aren't related to each other.


achille-roussel:
Thanks for the review!

PR #69, opened 2017-09-15
Use mutex instead of atomic operations

achille-roussel:
This PR makes to main changes:
- Change the use of atomic operations in favor of mutexes, testing with Go 1.9 there's almost no different anymore, definitely good improvements have been made to the mutex implementation, so I'd rather have an implementation that's easier to read and reason about.
- Don't spawn new goroutines to flush the measures. This was actually causing pretty high contention on the goroutine scheduler on a very high traffic service that was producing a lot of metrics.

### DIFF
--- a/buffer.go
+++ b/buffer.go
@@ -3,9 +3,7 @@ package stats
 import (
 	"io"
 	"sync"
-	"sync/atomic"
 	"time"
-	"unsafe"
 )
 
 // Buffer is the implementation of a measure handler which uses a Serializer to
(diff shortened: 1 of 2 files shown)


buffer.go
```diff
@@ -28,114 +26,71 @@ type Buffer struct {
...
-			b.waitReleaseRLock()
-			h.join.Add(1)
-			go h.writeBuffer(b, int(n-l))
+			flush.writeTo(b.Serializer)
+			releaseBuffer(flush)
 		}
 	}
 
-	chunkPool.Put(c)
+	releaseChunk(chunk)
```
f2prateek: could we `defer` this after the `acquire`?

achille-roussel: `defer` is still not the cheapest construct in Go (it often shows up on performance profiles), in a critical code path like this one I'd rather not use it.

PR #74, opened 2017-10-30
added an option to add the request path as a tag

dscrobonia:
Creates a new constructor for Handler that will include the request path as a DataDog tag. Also removes filter for the `http_req_path` tag that was used to limit requests to DataDog.

### DIFF
--- a/datadog/append.go
+++ b/datadog/append.go
@@ -33,14 +33,6 @@ func appendMetric(b []byte, m Metric) []byte {
 
 func appendTags(b []byte, tags []stats.Tag) []byte {
 	for i, t := range tags {
-		if t.Name == "http_req_path" {
-			// Datadog has complained numerous times that the request paths
-			// generate too many custom metrics on their side, for now we'll
-			// simply strip it out until we can come up with a better strategy
-			// for handling those.
-			continue
-		}
-
 		if i != 0 {
 			b = append(b, ',')
 		}
--- a/datadog/client.go
+++ b/datadog/client.go
@@ -21,15 +21,17 @@ const (
 	// datadog.
 	DefaultBufferSize = 1024
 
-	// DefaultFilter is the default tag to filter before sending to
-	// datadog. Using the request path as a tag can overwhelm datadog's
-	// servers if there are too many unique routes due to unique IDs being a
-	// part of the path. Only change the default filter if there is a static
-	// number of routes.
-	DefaultFilter = "http_req_path"
-
 	// MaxBufferSize is a hard-limit on the max size of the datagram buffer.
 	MaxBufferSize = 65507
+)
+
+// DefaultFilter is the default tag to filter before sending to
+// datadog. Using the request path as a tag can overwhelm datadog's
+// servers if there are too many unique routes due to unique IDs being a
+// part of the path. Only change the default filter if there is a static
+// number of routes.
+var (
+	DefaultFilters = []string{"http_req_path"}
 )
 
 // The ClientConfig type is used to configure datadog clients.
@@ -40,7 +42,7 @@ type ClientConfig struct {
 	// Maximum size of batch of events sent to datadog.
 	BufferSize int
 
-	// List of tags to filter
+	// List of tags to filter. If left nil is set to DefaultFilters.
 	Filters []string
 }
 
@@ -72,9 +74,7 @@ func NewClientWith(config ClientConfig) *Client {
 	}
 
 	if config.Filters == nil {
-		config.Filters = []string{
-			DefaultFilter,
-		}
+		config.Filters = DefaultFilters
 	}
 
 	// transform filters from array to map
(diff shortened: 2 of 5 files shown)


httpstats/metrics.go
```diff
@@ -197,6 +202,15 @@ func (m *metrics) observeRequest(req *http.Request, op string, bodyLen int) {
 	m.http.method = req.Method
 	m.http.protocol = req.Proto
 	m.http.transferEncoding = transferEncoding
+
+	if m.options.pathObserved {
+		path := req.URL.Path
+
```
dscrobonia: the [docs](https://golang.org/pkg/net/url/) for URL.Path mention that sometimes it includes the leading slash and sometimes not. Added this for consistency

datadog/append.go
```diff
@@ -33,14 +33,6 @@ func appendMetric(b []byte, m Metric) []byte {
 
 func appendTags(b []byte, tags []stats.Tag) []byte {
 	for i, t := range tags {
-		if t.Name == "http_req_path" {
```
dscrobonia: @achille-roussel this is the filter I wasn't sure we should remove. If you know it's not being used anymore we can clean it up. But I don't want to take this out and have a flood of requests flow through that were being held back.

httpstats/metrics.go
```diff
@@ -197,6 +202,15 @@ func (m *metrics) observeRequest(req *http.Request, op string, bodyLen int) {
...
+		if strings.HasPrefix(path, "/") {
```
achille-roussel: How about doing it the other way around, and adding the leading `/` when it's missing? Seems like it's more natural to think of paths for HTTP requests as starting with a `/`.

dscrobonia:
Adding the tag by default and controlling the usage of it by filtering at the datadog client seems like it would be much cleaner & easier to use. It isn't immediately obvious to me how I would exactly add that, but I'll take a closer look in a little bit.

httpstats/metrics.go
dscrobonia: Seemed like it my be redundant when trying to filter by path in datadog and always adding the /. But cool by me.

achille-roussel:
The datadog client is configured with this data structure https://github.com/segmentio/stats/blob/master/datadog/client.go#L29-L35, the idea would be to add the filtering list there, and default to filter http_req_path to keep the behavior backward compatible.

dscrobonia:
So I can add a filter list to the ClientConfig, which would then need to pass it to the serializer, where it would be used to filter out tags from the Measure before calling [AppendMeasure](https://github.com/segmentio/stats/blob/master/datadog/client.go#L107), yeah? 
I was just thinking of using an array of strings in the client configs to add the tag names to, any reason for a map of structs?

achille-roussel:
Yep that's the idea 👍 

A list of strings is fine in the configuration and the syntax will be more user-friendly, but we should really convert it to a map internally because the lookup time will be O(1) instead of O(N) otherwise.

`map[string]struct{}` is the common way to represent a set of unique values in Go.

datadog/client.go
```diff
@@ -32,6 +39,9 @@ type ClientConfig struct {
 
 	// Maximum size of batch of events sent to datadog.
 	BufferSize int
+
+	// List of tags to filter
```
achille-roussel: It seems worth documenting that when left `nil` the default filters are used.

datadog/client.go
```diff
@@ -21,6 +21,13 @@ const (
 	// datadog.
 	DefaultBufferSize = 1024
 
+	// DefaultFilter is the default tag to filter before sending to
+	// datadog. Using the request path as a tag can overwhelm datadog's
+	// servers if there are too many unique routes due to unique IDs being a
+	// part of the path. Only change the default filter if there is a static
+	// number of routes.
+	DefaultFilter = "http_req_path"
```
achille-roussel: How about making this a slice:
```go
var (
    DefaultFilters = []string{"http_req_path"}
)
```
That would let programs potentially overwrite it, but also would be consistent to have the default be of the same type as the `Filters` field.

datadog/measure.go
```diff
@@ -9,7 +9,7 @@ import (
 
 // AppendMeasure is a formatting routine to append the dogstatsd protocol
 // representation of a measure to a memory buffer.
-func AppendMeasure(b []byte, m stats.Measure) []byte {
+func AppendMeasure(b []byte, m stats.Measure, filters map[string]struct{}) []byte {
```
achille-roussel: To avoid the breaking change here, do you think we could keep this function the way it was an introduce a new one instead?
```go
func AppendMeasure(b []byte, m stats.Measure) []byte {
    return appendMeasureFiltered(b, m, nil)
}

func appendMeasureFiltered(b []byte, m stats.Measure, filters map[string]struct{}) []byte {
    // ...
}
```

datadog/client.go
dscrobonia: Ah okay nice. I saw I couldn't add slices/arrays as constants, which is why I did it the way I did. Much better!

achille-roussel: Yep you can't make it a constant, but you can make it a top-level variable, this is common practice.

datadog/measure.go
dscrobonia: Ya definitely. A quick search of the repo gave me the impression it was only used in the one spot, which is why I went for the change.

achille-roussel: Yes, it's very unlikely that there is code out there that depends on this function, but since it's exported we have no way to know.

PR #77, opened 2017-11-15
cleanup procstats

achille-roussel:
Quick cleanup:
- Fix a typo
- Don't collect process stats right before stopping, it causes inflated CPU usage to be reported
- Collect CPU stats every 15 seconds, doing it every 5 seconds also causes undesirable spiky stats


achille-roussel:
Actually I was wrong, the misreported CPU usage wasn't happening on stop, it was happening on start, the latest commit addresses it.

PR #78, opened 2017-12-10
global measure cache

achille-roussel:
The measure cache was previously kept in the engine, which had a few disadvantages:
- it caused escape analysis to always put the engine value on the heap, because internally stats.(*Engine).ReportAt would take the address of the cache, making the creation of short-lived engines quite expensive
- new engines would not share the cache from the engine they were created from, which means they would have to rebuild it on their own, making the program slower when creating short-lived engines

This PR changes the implementation to use a global cache instead, which is shared by all stat engines.


achille-roussel:
Nevermind I realized we can't do this because it actually caches the engine prefix. I'll leave a note instead.

PR #83, opened 2018-04-11
WriteStats in prometheus handler

rohitggarg:
For those who are not using HTTP and still need to write stats through io.Writer

### DIFF
--- a/prometheus/handler.go
+++ b/prometheus/handler.go
@@ -122,9 +122,6 @@ func (h *Handler) ServeHTTP(res http.ResponseWriter, req *http.Request) {
 		return
 	}
 
-	metrics := h.metrics.collect(make([]metric, 0, 10000))
-	sort.Sort(byNameAndLabels(metrics))
-
 	w := io.Writer(res)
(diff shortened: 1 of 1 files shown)


prometheus/handler.go
```diff
@@ -135,9 +132,16 @@ func (h *Handler) ServeHTTP(res http.ResponseWriter, req *http.Request) {
 		w = zw
 	}
 
+	h.WriteStats(w)
+}
+
+func (h *Handler) WriteStats(w io.Writer) {
```
achille-roussel: Would you mind adding a bit of documentation to this method?

rohitggarg: Thanks!!
I will put some documentation tomorrow and do a bump

rohitggarg:
Hi, I have added a basic doc. Is there like a guideline on how to contribute to documentation in this repository?

achille-roussel:
We don't have guidelines, but your contribution looks great 👍 

Thanks for the PR, I'll go ahead and merge!

PR #85, opened 2018-07-29
Procstats fix cpu

achille-roussel:
This PR implements proper collection of CPU time consumed by processes other than the program itself.

### DIFF
--- a/procstats/proc_linux.go
+++ b/procstats/proc_linux.go
@@ -1,24 +1,61 @@
 package procstats
 
 import (
+	"io/ioutil"
 	"os"
+	"strconv"
+	"strings"
+	"sync"
 	"syscall"
 	"time"
 
 	"github.com/segmentio/stats/procstats/linux"
 )
 
+var (
+	clockTickOnce  sync.Once
+	clockTickValue uint64
+	clockTickError error
+)
+
+func clockTick() (uint64, error) {
+	s, err := getconf("CLK_TCK")
+	if err != nil {
+		return 0, err
+	}
+	return strconv.ParseUint(strings.TrimSpace(s), 10)
+}
+
+func getconf(name string) (string, error) {
+	file, err := exec.LookupPath("getconf")
+	if err != nil {
+		return 0, err
+	}
+
+	r, w, err := os.Pipe()
+	if err != nil {
+		return 0, nil
+	}
+	defer r.Close()
+	defer w.Close()
+
+	p, err := os.StartProcess(file, []string{"getconf", name}, &os.ProcAttr{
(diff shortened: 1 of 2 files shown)


procstats/proc_linux.go
```diff
@@ -1,25 +1,70 @@
...
+	clockTickHertz uint64
+	clockTickError error
+)
 
-	pagesize := uint64(syscall.Getpagesize())
-	rusage := syscall.Rusage{}
+func clockTicksToDuration(ticks uint64) time.Duration {
+	clockTickOnce.Do(func() { clockTickHertz, clockTickError = clockTick() })
+	check(clockTickError)
+	return time.Duration(1e9 * float64(ticks) / float64(clockTickHertz))
```
jnjackins: Why not use `time.Second` instead of `1e9`?

procstats/proc_linux.go
```diff
@@ -1,25 +1,70 @@
...
+}
 
-	if pid == os.Getpid() {
-		check(syscall.Getrusage(syscall.RUSAGE_SELF, &rusage))
-	} else {
-		// TODO: figure out how to get accurate stats for other processes
+func clockTick() (uint64, error) {
+	s, err := getconf("CLK_TCK")
```
jnjackins: Is this new configuration?

jnjackins:
Are we sure we want to lump the children CPU times in with the parent? I think it might be a bit surprising in some cases. Also, I think it's possible the child CPU times won't always be accurate (ticks per second can change during execution, with dynamic frequency scaling). Would it be better to have a separate metric for CPU time of child procs?

procstats/proc_linux.go
achille-roussel: `time.Second` is a `time.Duration`, but the expression here works with floats. It would have to be `float64(time.Second)` which is much more verbose.

I can change it if you feel strongly about it.

achille-roussel:
Tick-per-second can't change, it's a constant built in the kernel (and glibc for the matter), it's independent of the actual frequency of the processor.

Maybe the code isn't really clear here, the parent/child metrics aren't mixed together. When collecting metrics the caller specifies the PID of the process it wants to get metrics for, if it's its own PID it gets metrics from its own process, but it can read metrics of any other process it has permissions for.

Let me know if I've overlooked something.

procstats/proc_linux.go
jnjackins: I don't feel that strongly :)

jnjackins:
Ahh, yep, I misunderstood. Looks good!

achille-roussel:
Thanks for the review!

PR #86, opened 2018-08-01
Adding veneur integration for stats

systemizer:
Note: this is a first pass. If we agree with this direction, I can go ahead and add tests.

Veneur [supports the DogStatsd protocol](https://github.com/stripe/veneur#what-is-veneur), so I chose to inherit the statsd functionality from the existing Datadog integration.

Veneur supports [Magic Tags](https://github.com/stripe/veneur#magic-tag) that help with routing metrics. We need this feature to route high cardinality metrics only to the Signalfx backend.

We need to add the `VeneurGlobalOnly` Tag for all SignalFX metrics or else our DPM will be higher than what our contract permits. If we set `VeneurGlobalOnly` on the Engine's Global tags, this will affect our current Datadog metrics (which is not something we want to do).

So, by adding veneur tags only to a veneur backend as I've done here, we can:
1. Register a Veneur Handler with VeneurGlobalOnly and enable only the SignalfxSink
2. Register a Datadog Handler (to keep metrics as they were)
3. Write metrics to both such that we both get high cardinality metrics in Signalfx and don't change the behavior of our current Datadog setup.

This is more complex than I would like it to be, but due to some limitations with routing metrics and feature imparity between Signalfx and Datadog, this is what I've come up with.

I'm open to other suggestions. Let me know what you think!

### DIFF
--- a/veneur/client.go
+++ b/veneur/client.go
@@ -8,9 +8,9 @@ import (
 )
 
 const (
-	TagVeneurGlobalOnly = "veneurglobalonly"
-	TagVeneurLocalOnly  = "veneurlocalonly"
-	TagVeneurSinkOnly   = "veneursinkonly"
+	GlobalOnly = "veneurglobalonly"
+	LocalOnly  = "veneurlocalonly"
+	SinkOnly   = "veneursinkonly"
 )
 
 // The ClientConfig type is used to configure datadog clients.
@@ -50,12 +50,12 @@ func NewClientWith(config ClientConfig) *Client {
 	// Construct Veneur-specific Tags we will append to measures
 	tags := []stats.Tag{}
 	if config.GlobalOnly {
-		tags = append(tags, stats.Tag{Name: TagVeneurGlobalOnly})
+		tags = append(tags, stats.Tag{Name: GlobalOnly})
 	} else if config.LocalOnly {
-		tags = append(tags, stats.Tag{Name: TagVeneurLocalOnly})
+		tags = append(tags, stats.Tag{Name: LocalOnly})
 	}
 	for _, t := range config.SinksOnly {
-		tags = append(tags, stats.Tag{Name: TagVeneurSinkOnly, Value: t})
+		tags = append(tags, stats.Tag{Name: SinkOnly, Value: t})
 	}
 
 	return &Client{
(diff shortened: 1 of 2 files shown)


veneur/client.go
```diff
@@ -0,0 +1,77 @@
...
+
+import (
+	"time"
+
+	"github.com/segmentio/stats"
+	"github.com/segmentio/stats/datadog"
+)
+
+const (
+	TagVeneurGlobalOnly = "veneurglobalonly"
```
achille-roussel: How about dropping the `TagVeneur` prefix here? When using this package, it'll be just as expressive and less verbose to write `veneur.GlobalOnly` than `veneur.TagVeneurGlobalOnly`.

veneur/client.go
```diff
@@ -0,0 +1,77 @@
...
+			BufferSize: config.BufferSize,
+			Filters:    config.Filters,
+		}),
+		tags: tags,
+	}
+}
+
+// HandleMetric satisfies the stats.Handler interface.
+func (c *Client) HandleMeasures(time time.Time, measures ...stats.Measure) {
+	for _, m := range measures {
```
achille-roussel: You should make a copy of the input slice, otherwise you're modifying the values received by this method and there's a chance that you'll be passing the tags to other handlers.

veneur/client.go
```diff
@@ -0,0 +1,77 @@
...
+		m.Tags = append(m.Tags, c.tags...)
```
achille-roussel: This should be `measures[i].Tags = ` shouldn't it? The `m` variable is not a reference to the slice items, a copy is made for each loop iteration.

boggsboggs:
I'm onboard with the setup here. Only one loose end to consider from the veneur Readme:
```Note: For global counters to report correctly, the local and global Veneur instances should be configured to have the same flush interval.```

Our current datadog veneur is configured to 10s, I think. And, signalFX is configured to 60s.

We might have to send the signalFX metrics straight to the veneur-proxy hosts, like you suggested earlier.

systemizer:
This is ready for review. @achille-roussel I decide not to add a tag-list to the current datadog client because global tags are currently configured at the engine level. IMO, it would only make sense to add tags at the handler level if there was specific tags that affected how metrics were handled by that particular handler (in this case, veneur uses magic tags). That being said, my feelings about the matter aren't very strong; added a global tag list to the datadog handler would add less code but might result in confusion in the future if some people add global tags to the handler while others add it to the engine

PR #87, opened 2018-08-09
Add process delay statistics collection for Linux

otterley:
Add process delay statistics collection capability on Linux.  This requires access to the taskstats Netlink generic family, which imposes a few limitations:

* If run in a container, it must be run with host networking enabled (e.g. `docker run --net=host`)
* The process must be run with at least `CAP_NET_RAW` capability

It's also currently dependent on https://github.com/mdlayher/taskstats/pull/5 which I hope will be merged soon.


procstats/delaystats.go
```diff
@@ -0,0 +1,54 @@
...
+// NewDelayStatsWith collects metrics on the process identified by pid and
+// reports them to eng.
+func NewDelayMetricsWith(eng *stats.Engine, pid int) *DelayMetrics {
+	p := &DelayMetrics{engine: eng, pid: pid}
+	return p
+}
+
+// Collect satisfies the Collector interface.
+func (d *DelayMetrics) Collect() {
+	if info, err := CollectDelayInfo(d.pid); err == nil {
```
rjenkins: Should we log err here or are we concerned about spamming logs?

procstats/delaystats_linux.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/segmentio/taskstats"
+)
+
+func collectDelayInfo(pid int) (info DelayInfo, err error) {
+	client, err := taskstats.New()
+	if err != nil {
+		if err1, ok := err.(syscall.Errno); ok && err1 == syscall.ENOENT {
+			panic(errors.New("Failed to communicate with taskstats Netlink family.  Ensure this program is not running in a network namespace."))
+		}
+		check(err)
```
rjenkins: Maybe just panic here and below as we already know err != nil?

procstats/delaystats_linux.go
```diff
@@ -0,0 +1,32 @@
...
+	}
+	stats, err := client.PID(pid)
+	if err != nil {
+		if err1, ok := err.(syscall.Errno); ok && err1 == syscall.EPERM {
+			panic(errors.New("Failed to open Netlink socket: permission denied.  Ensure CAP_NET_RAW is enabled for this process, or run it with root privileges."))
+		}
+		check(err)
```
rjenkins: Same as above, just panic?

procstats/delaystats_linux.go
```diff
@@ -0,0 +1,32 @@
...
+	"errors"
+	"syscall"
+
+	"github.com/segmentio/taskstats"
+)
+
+func collectDelayInfo(pid int) (info DelayInfo, err error) {
+	client, err := taskstats.New()
+	if err != nil {
+		if err1, ok := err.(syscall.Errno); ok && err1 == syscall.ENOENT {
```
achille-roussel: This type check is unnecessary, you could just check for `if err == syscall.ENOENT`

procstats/delaystats_linux.go
```diff
@@ -0,0 +1,32 @@
+package procstats
+
+import (
+	"errors"
+	"syscall"
+
+	"github.com/segmentio/taskstats"
```
achille-roussel: What motivated to bring this in-house instead of taking a dependency to another repository?

procstats/delaystats.go
```diff
@@ -0,0 +1,54 @@
...
+// the default stats engine.
+func NewDelayMetrics() *DelayMetrics {
+	return NewDelayMetricsWith(stats.DefaultEngine, os.Getpid())
+}
+
+// NewDelayStatsWith collects metrics on the process identified by pid and
+// reports them to eng.
+func NewDelayMetricsWith(eng *stats.Engine, pid int) *DelayMetrics {
+	p := &DelayMetrics{engine: eng, pid: pid}
+	return p
```
achille-roussel: How about removing this local variable here?

otterley:
> Question on the design: Why not include the collection of delay metrics in the process stats that already existed? For example adding Delay time.Duration to the CPUInfo data structure.

Good question!  It was considered, but the fact that collection of these stats requires elevated privileges and host networking means that a number (most?) of the programs we run won't be able to collect them, unless we significantly change their configuration.

Another option would be to collect them if it's possible by default, and maybe emit a warning if we can't.  Would you prefer that?

otterley:
Applied fixes per feedback.

procstats/delaystats_linux.go
otterley: It's currently dependent on a PR that's waiting to be merged.  Once it's merged we can change this to be the vendor repository again.

PR #90, opened 2018-09-27
fix typo in splitMeasureField()

yerden:
Current `splitMeasureField()` implementation leads to non-equivalent behaviour of old-fashioned `stats.Incr()` and new `stats.Report()` APIs.

### DIFF
--- a/buckets.go
+++ b/buckets.go
@@ -34,10 +34,10 @@ func makeKey(s string) Key {
 }
 
 func splitMeasureField(s string) (measure string, field string) {
-	if i := strings.LastIndexByte(s, ':'); i >= 0 {
+	if i := strings.LastIndexByte(s, '.'); i >= 0 {
 		measure, field = s[:i], s[i+1:]
 	} else {
-		measure = s
+		field = s
 	}
 	return
 }
--- a/engine_test.go
+++ b/engine_test.go
@@ -128,13 +128,13 @@ func testEngineIncr(t *testing.T, eng *stats.Engine) {
 
 	checkMeasuresEqual(t, eng,
 		stats.Measure{
-			Name:   "test.measure.count",
-			Fields: []stats.Field{stats.MakeField("", 1, stats.Counter)},
+			Name:   "test.measure",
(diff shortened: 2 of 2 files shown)


yerden:
This may be proven by this piece of code:
``` go
package main

import (
	"flag"
	"github.com/segmentio/stats"
	"github.com/segmentio/stats/prometheus"
	"net/http"
)

type myMetrics struct {
	dice struct {
		vertex struct {
			rolls  int `metric:"rolls" type:"counter"`
			throws int `metric:"throws" type:"counter"`
		} `metric:"vertex"`
	} `metric:"dice"`
}

var useReport = flag.Bool("use-report", false, "Use report instead of Add()")

func main() {
	flag.Parse()

	handler := prometheus.DefaultHandler
	stats.Register(handler)
	http.Handle("/metrics", handler)

	go func() {
		if *useReport {
			m := &myMetrics{}
			m.dice.vertex.rolls = 1
			m.dice.vertex.throws = 1
			stats.Report(m)
		} else {
			stats.Incr("dice.vertex.rolls")
			stats.Incr("dice.vertex.throws")
		}
	}()

	http.ListenAndServe(":8080", nil)
}
```

And here we go:
``` bash
$ go run main.go &
[1] 31710
$ curl localhost:8080/metrics
dice_vertex_rolls_ 1 1538064026104
dice_vertex_throws_ 1 1538064026104
$ kill %1
$ go run main.go -use-report &
[2] 31798
[1]   Terminated              go run main.go
$ curl localhost:8080/metrics
# TYPE dice_vertex_rolls counter
dice_vertex_rolls 1 1538064037729

# TYPE dice_vertex_throws counter
dice_vertex_throws 1 1538064037729
```

yerden:
oops, test is failing...

yerden:
That's it. Tested some corner cases. Looks like the code behaves the same way with both API flavours now.

yerden:
any news here?

achille-roussel:
I was a bit concerned that this may alter the behavior of some other cases in subtle ways, but after doing a bit more testing I didn't notice any unexpected changes.

Thanks for the contribution @yerden

PR #96, opened 2018-11-19
normalize CPU usage metrics

achille-roussel:
This PR adds the CPU period, quota, and shares to the info reported by the procstats package, and uses it to normalize the CPU usage metrics.

I've also spent a bit of time cleaning up parts of the code, this package was almost two years old and deserved a bit of love.

### DIFF
--- a/procstats/proc.go
+++ b/procstats/proc.go
@@ -133,16 +133,25 @@ func (p *ProcMetrics) Collect() {
 		now := time.Now()
 
 		if !p.lastTime.IsZero() {
-			interval := now.Sub(p.lastTime)
+			ratio := 1.0
+			switch {
+			case m.CPU.Period > 0 && m.CPU.Quota > 0:
+				ratio = float64(m.CPU.Quota) / float64(m.CPU.Period)
+			case m.CPU.Shares > 0:
+				ratio = float64(m.CPU.Shares) / 1024
+			}
+
+			interval := ratio * float64(now.Sub(p.lastTime))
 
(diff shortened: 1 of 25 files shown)


procstats/proc.go
```diff
@@ -133,16 +133,25 @@ func (p *ProcMetrics) Collect() {
 		now := time.Now()
 
 		if !p.lastTime.IsZero() {
-			interval := now.Sub(p.lastTime)
+			ratio := 1.0
```
Pryz: I'm probably mis-understanding something here. But it seems that `ratio` will be the number of cores if we are using a quota or shares. Should we be using the number of CPUs for the default use case ?

procstats/proc.go
```diff
@@ -182,6 +191,13 @@ func CollectProcInfo(pid int) (ProcInfo, error) {
 type CPUInfo struct {
 	User time.Duration // user cpu time used by the process
 	Sys  time.Duration // system cpu time used by the process
+
+	// Linux-specific details about the CPU configuration of the process.
```
Pryz: Maybe we could add a reference to https://www.kernel.org/doc/Documentation/scheduler/sched-bwc.txt to provide the reader more doc ?

procstats/proc.go
achille-roussel: That's a good idea. I'm not sure what's more _natural_ to reason about, for example having 100% be 1 core, and when it's using more it goes over 100 (kind of like top does it), or normalizing based on the number of cores available.

Issue #93, opened 2018-11-15
Per request tags


mnichols:
Hi we are using the `httpstats` Transport decorator like this:

```
	tags := []stats.Tag{
                // this tag is OK to always be the same value
		{Name: TagCallingService, Value: opts.CallingService},
                // but this tag should be determined by some properties of the outbound request
		{Name: TagRequestName, Value: computeRequestName()},
	}

	return httpstats.NewTransportWith(stats.DefaultEngine.WithTags(tags...), inner)
```

However, we need to provide a couple of tags with values determined by the request (based on the path, etc). It seems to be bad practice to create a new client / transport per request but the `httpstats` Transport decorator doesnt seem to provide a way to add a couple of tags during a request.
Is there some kind of callback or hook for adding tags that are scoped to a request?

abraithwaite:
Fixed by #104 and #111.

PR #100, opened 2019-01-28
Add support for per request tags

maxcnunes:
As reported on https://github.com/segmentio/stats/issues/93 and documented in Go api it seems to be a bad practice using a new client/transport instance on every request.

https://golang.org/pkg/net/http/#Client
> The Client's Transport typically has internal state (cached TCP connections), so Clients should be reused instead of created as needed. Clients are safe for concurrent use by multiple goroutines.

and

> Clients and Transports are safe for concurrent use by multiple goroutines and for efficiency should only be created once and re-used.

So, a solution that came to my mind is passing the tags by Context and creating the stats engine with thos provided tags during the round trip.

I would like to get a feedback if segmentio team is interested on this changes before adding the proper documentation and tests for this change.

### DIFF
--- a/httpstats/transport.go
+++ b/httpstats/transport.go
@@ -7,6 +7,8 @@ import (
 	"github.com/segmentio/stats"
 )
 
+const contextKeyReqTags = "segmentio_httpstats_req_tags"
+
 // NewTransport wraps t to produce metrics on the default engine for every request
 // sent and every response received.
 func NewTransport(t http.RoundTripper) http.RoundTripper {
@@ -27,12 +29,26 @@ type transport struct {
 	eng       *stats.Engine
 }
 
+// RequestWithContext returns a shallow copy of req with its context changed with this provided tags
+// so the they can be used later during the RoundTrip in the metrics recording.
+// The provided ctx must be non-nil.
+func RequestWithContext(req *http.Request, tags ...Tags) *http.Request {
+	ctx := req.Context()
+	ctx = ctx.WithValue(ctx, contextKeyReqTags, tags)
+	return req.WithContext(ctx)
+}
+
 func (t *transport) RoundTrip(req *http.Request) (res *http.Response, err error) {
 	start := time.Now()
 	rtrip := t.transport
+	eng := t.eng
 
 	if rtrip == nil {
 		rtrip = http.DefaultTransport
+	}
+
+	if tags, ok := req.Context().Value(contextKeyReqTags).([]stats.Tag); ok {
+		eng = eng.WithTags(tags...)
 	}
 
 	if req.Body == nil {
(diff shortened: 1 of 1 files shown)


httpstats/transport.go
```diff
@@ -27,22 +29,36 @@ type transport struct {
...
 	start := time.Now()
 	rtrip := t.transport
+	eng := t.eng
 
 	if rtrip == nil {
 		rtrip = http.DefaultTransport
 	}
 
+	if tags, ok := req.Context().Value(contextKeyReqTags).([]stats.Tag); ok {
+		eng = eng.WithTags(tags...)
```
maxcnunes: Wondering if creating a new copy of Engine per request will have a high cpu/memory cost

maxcnunes:
cc @achille-roussel @rjenkins

maxcnunes:
hey @achille-roussel any news on this? I was thinking, another approach would be to extract that logic in the RoadTrip into something like a span which we call before running the request and providing the request instance as an argument and after the response had been executed and then providing the response instance to it. It would require more steps to use it, but at least it would provide an API to report an outgoing request metric. Something like this:
```go
span := httpstats.StartOutgoingRequestSpan(req)
res, err := client.Do(req)
span.Finish(res)
```

Also, I'm worried what is the memory cost for creating a new engine with`WithTags` for every request just so I can add tags such as `endpoint`, `request_source`, `direction:outgoing`. Wouldn't be more eficient if we made requestBody and responseBody public and they had `meta interface{}` field so we could add custom tags to it using the field tags?

maxcnunes:
Btw, I willing to help implementing any solution :)

adamdicarlo:
@achille-roussel have you had a chance to give this some thought?

httpstats/transport.go
achille-roussel: It will add a couple of heap-allocations, but since it's an opt-in feature I think that's fine.

httpstats/transport.go
```diff
@@ -27,22 +29,36 @@ type transport struct {
 	eng       *stats.Engine
 }
 
+// RequestWithContext returns a shallow copy of req with its context changed with this provided tags
+// so the they can be used later during the RoundTrip in the metrics recording.
+// The provided ctx must be non-nil.
+func RequestWithContext(req *http.Request, tags ...Tags) *http.Request {
```
achille-roussel:
```suggestion
func RequestWithTags(req *http.Request, tags ...Tags) *http.Request {
```

httpstats/transport.go
```diff
@@ -7,6 +7,8 @@ import (
 	"github.com/segmentio/stats"
 )
 
+const contextKeyReqTags = "segmentio_httpstats_req_tags"
```
achille-roussel: How about replacing this with an internal empty type to avoid any risk of conflicts? For example
```go
type tagsKey struct{}
```

mnichols:
Hi I reissued this PR from the fork we used at InVisionApp's repo: https://github.com/segmentio/stats/pull/104.
It was updated with these changes.

abraithwaite:
Did #104 work for you over at InvisionApp @mnichols @maxcnunes ?  Can we close this?

abraithwaite:
Going to close this for now.  I think both #104 and #111 together should satisfy this use case.  Thanks!

PR #102, opened 2019-03-18
add support for metric arrays

achille-roussel:
This PR extends the package to support reporting array of metrics instead of just structs.


achille-roussel:
Thanks for the review!

PR #104, opened 2019-04-01
Add support for tags per request: InVisionApp fork

mnichols:
This implements the changes requested in the PR [here](https://github.com/segmentio/stats/pull/100) but from our org repo. 
Thanks for the feedback!

### DIFF
--- a/httpstats/transport.go
+++ b/httpstats/transport.go
@@ -7,6 +7,8 @@ import (
 	"github.com/segmentio/stats"
 )
 
+const contextKeyReqTags = "segmentio_httpstats_req_tags"
+
 // NewTransport wraps t to produce metrics on the default engine for every request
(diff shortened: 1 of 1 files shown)


go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/segmentio/stats
+module github.com/InVisionApp/stats
```
maxcnunes: @mnichols it looks like you need to revert this part

mnichols: ah nice catch...fixed! @maxcnunes

achille-roussel:
Thanks for the contribution!

mnichols:
You're welcome! Here is a PR that has tests and docs for this ... you merged it before I could update the branch 😄 . https://github.com/segmentio/stats/pull/105

PR #105, opened 2019-04-02
added tests and docs for new transport facilities

mnichols:
This adds documentation for `RequestWithTags` func and also tests to verify the custom tags are sent out.

### DIFF
--- a/README.md
+++ b/README.md
@@ -315,6 +315,30 @@ func main() {
 }
 ```
 
+You may also add specific tags per request.
+```go
+package main
+
+import (
+    "net/http"
+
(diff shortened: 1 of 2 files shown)


mnichols:
Looks like request stats dont get reported home from a wrapped transport. I think `requestBody` is missing this `r.eng.Report(metrics)` in its `complete` func, like the `responseBody` has 

```
func (r *responseBody) complete() {
	r.metrics.observeResponse(r.res, r.op, r.bytes, time.Now().Sub(r.start))
	r.eng.ReportAt(r.start, r.metrics)
}

```

PR #107, opened 2019-04-29
tag: add json struct tags

abraithwaite:
While not strictly necessary, these would be nice to have.  Currently
pulling in these tags to log in my app as well and would rather not loop
over them to marshal them with "proper" (lowercase) json tags.


abraithwaite:
I actually don't like this strategy.  I updated my local code to iterate over it anyway to have a clean tag map: `map[string]string` in JSON.

PR #108, opened 2019-08-12
fix netstat metric names

achille-roussel:
Follow up of #90 which fixed a backward incompatible change, but kind of broke the (rarely used) API...

We could have done a better job managing major versions with this package, not sure if this should represent a major version bump or not. Let me know what you think!

### DIFF
--- a/netstats/conn.go
+++ b/netstats/conn.go
@@ -42,7 +42,7 @@ func NewConnWith(eng *stats.Engine, c net.Conn) net.Conn {
 	nc.r.metrics.protocol = proto
 	nc.w.metrics.protocol = proto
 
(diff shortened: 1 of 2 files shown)


abraithwaite:
:man_shrugging:  I don't have a strong opinion on it . How'd it break the API?  Maybe that can inform whether or not we bump the version.

achille-roussel:
We supported metric names with `<metric name>:<field name>`, now the model is `<metric name>.<field name>` where the last dot is assumed to be the start of the field name.

I doubt anyone depended on this because it wasn't documented 🤷‍♂

abraithwaite:
:shipit:

PR #109, opened 2019-09-11
Add user agent tag for httpstats package

systemizer:
Curious whether you've thought or would be willing to add user agent as a tag for http stats. It's unbounded in the number of permutations so it could result in high cardinality. At the same time, for internal services, where user agents are generally constrained, it would be very helpful. Could also update the PR to add configuration such that user agent is by default disabled, but give the user the ability to enable it for specific services

### DIFF
--- a/httpstats/handler.go
+++ b/httpstats/handler.go
@@ -21,7 +21,7 @@ func NewHandlerWith(eng *stats.Engine, h http.Handler) http.Handler {
 	return &handler{
 		handler: h,
 		eng:     eng,
-		config:  defaultHandlerConfig,
+		conf:    defaultHandlerConfig,
 	}
 }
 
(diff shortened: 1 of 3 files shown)


Pryz:
Should we make this optional to avoid blowing up the cardinality ?

systemizer:
Yeah im up for that (mentioned that in the description). I can make it configurable

systemizer:
@Pryz this is ready for review. no rush

httpstats/handler.go
```diff
@@ -118,5 +140,12 @@ func (w *responseWriter) complete() {
 	}
 
 	w.metrics.observeResponse(res, "write", w.bytes, now.Sub(w.start))
+
+	// UserAgent can result in high cardinality. Allow mechanism for disabling the tag
+	if w.conf.DisableUserAgent {
+		w.metrics.http.userAgent = ""
+
```
Pryz: nit

systemizer: sorry, what's the nit you're referring to?

f2prateek:
One thing I think could be helpful is we had a nicer API to augment the default stats emitted by this package, instead of manually adding options for things like path and userAgent. Something like this https://github.com/segmentio/stats/pull/100/files, that relies on the context for instance.

f2prateek:
If we do go with this approach, I think for the sake of API consistency, it might be better instead to use the `DefaultFilters` option like we do for `http_path` - https://github.com/segmentio/stats/blob/17e5e763373e9d6a80804ff4dab85a842ad03a09/datadog/client.go#L34 (clients can disable it like this https://github.com/segmentio/ctlstore/pull/14/files).

abraithwaite:
This is possible now with #111

PR #110, opened 2019-09-17
Prometheus Handler: add the ability to ignore labels by name

schigh:
## TL;DR:
This PR is based on the need to exclude certain high-cardinality labels from the prometheus handler when collecting stats.  
In our specific use case, we needed to filter out the `http_req_path` label (from github.com/segmentio/stats/httpstats) because many of our URIs contain UUIDs and the like.

## Usage:
```go
// create prometheus handler
handler := &prometheus.Handler{
    TrimPrefix: "my_service_name",
}
// set HC tags
handler.IgnoreLabels([]string{"http_req_path"})
```

## Benchmarks:

`go test -bench BenchmarkHandleMetric -benchmem`

`current master`:
```
goos: darwin
goarch: amd64
pkg: github.com/segmentio/stats/prometheus
BenchmarkHandleMetric/counter-8                  5687409               189 ns/op               0 B/op          0 allocs/op
BenchmarkHandleMetric/gauge-8                    6038472               187 ns/op               0 B/op          0 allocs/op
BenchmarkHandleMetric/histogram-8                5275975               230 ns/op               0 B/op          0 allocs/op
PASS
ok      github.com/segmentio/stats/prometheus   4.078s
```
`this branch`:
```
goos: darwin
goarch: amd64
pkg: github.com/segmentio/stats/prometheus
BenchmarkHandleMetric/counter-8                  6561445               186 ns/op               0 B/op          0 allocs/op
BenchmarkHandleMetric/gauge-8                    6449792               184 ns/op               0 B/op          0 allocs/op
BenchmarkHandleMetric/histogram-8                5353644               223 ns/op               0 B/op          0 allocs/op
PASS
ok      github.com/segmentio/stats/prometheus   4.224s
```


prometheus/label.go
```diff
@@ -17,12 +19,12 @@ func (l label) hash() uint64 {
 	return h
 }
 
-func (l1 label) equal(l2 label) bool {
```
schigh: This was done to make the receiver names consistent.  It is not part of the logic concerning this PR.  I can reverse these chunks if there is any objection.

prometheus/handler.go
```diff
@@ -45,8 +46,21 @@ type Handler struct {
...
+// IgnoreLabels will remove the passed labels from the observation
+func (h *Handler) IgnoreLabels(labelNames []string) {
+	b := make([][]byte, len(labelNames))
+	for i := range labelNames {
+		b[i] = []byte(labelNames[i])
+	}
+	// This flattens all the ignored labels into a single byte slice
+	// so we dont have to do an explicit O(n^2) lookup when we filter
+	// labels out of the observation
+	h.ignoredLabels = bytes.Join(b, []byte{0x00})
```
abraithwaite: Hmm, this seems unnecessary.  Whether or not you're iterating over a slice of strings or doing `bytes.Contains` shouldn't really have a significant performance difference, and the former is much more clear.  They're both `O(n*m)` where n is the number of tags and m is the number of ignored, not `O(n**2)`.

If we add this, I would suggest using a list of strings representing the labels instead.

schigh: Thanks @abraithwaite.  Your suggested changes have been submitted

PR #111, opened 2019-09-19
context: add support for carrying tags on context, use in httpstats.Handler

abraithwaite:
Recently had a need to add this, and I can see it being useful generally.

It _might_ supersede #109, but I could see them both being useful and not exclusive of one another.

In fact, @f2prateek suggested exactly this, haha: https://github.com/segmentio/stats/pull/109#issuecomment-530580543

### DIFF
--- a/context.go
+++ b/context.go
@@ -3,12 +3,16 @@ package stats
 import "context"
 
 func ContextWithTags(ctx context.Context, tags ...Tag) context.Context {
-	if x := getTagSlice(ctx); x != nil {
-		*x = append(*x, tags...)
-		return ctx
-	}
 	// initialize the context reference and return a new context
 	return context.WithValue(ctx, contextKeyReqTags, &tags)
+}
+
+func ContextAddTags(ctx context.Context, tags ...Tag) bool {
+	if x := getTagSlice(ctx); x != nil {
+		*x = append(*x, tags...)
+		return true
+	}
+	return false
 }
 
 func ContextTags(ctx context.Context) []Tag {
--- a/context.go
+++ b/context.go
@@ -2,11 +2,21 @@ package stats
 
 import "context"
 
+// ContextWithTags returns a new child context with the given tags.  If the
+// parent context already has tags set on it, they are _not_ propegated into
+// the context children.
 func ContextWithTags(ctx context.Context, tags ...Tag) context.Context {
 	// initialize the context reference and return a new context
 	return context.WithValue(ctx, contextKeyReqTags, &tags)
 }
 
+// ContextAddTags adds the given tags to the given context, if the tags have
+// been set on any of the ancestor contexts.  ContextAddTags returns true
+// if tags were successfully appended to the context, and false otherwise.
+//
+// The proper way to set tags on a context if you don't know whether or not
+// tags already exist on the context is to first call ContextAddTags, and if
+// that returns false, then call ContextWithTags instead.
 func ContextAddTags(ctx context.Context, tags ...Tag) bool {
 	if x := getTagSlice(ctx); x != nil {
 		*x = append(*x, tags...)
@@ -15,6 +25,8 @@ func ContextAddTags(ctx context.Context, tags ...Tag) bool {
 	return false
 }
 
+// ContextTags returns a copy of the tags on the context if they exist and nil
+// if they don't exist
 func ContextTags(ctx context.Context) []Tag {
 	if x := getTagSlice(ctx); x != nil {
 		ret := make([]Tag, len(*x))
--- a/context.go
+++ b/context.go
@@ -1,13 +1,18 @@
 package stats
 
-import "context"
+import (
+	"context"
+	"sync"
+)
 
 // ContextWithTags returns a new child context with the given tags.  If the
 // parent context already has tags set on it, they are _not_ propegated into
 // the context children.
 func ContextWithTags(ctx context.Context, tags ...Tag) context.Context {
 	// initialize the context reference and return a new context
-	return context.WithValue(ctx, contextKeyReqTags, &tags)
+	return context.WithValue(ctx, contextKeyReqTags, &tagSlice{
+		tags: tags,
+	})
 }
 
 // ContextAddTags adds the given tags to the given context, if the tags have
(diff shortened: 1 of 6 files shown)


abraithwaite:
If we like this, I'll add more thorough tests which exercise the handler path.

httpstats/context.go
```diff
@@ -0,0 +1,62 @@
...
+	"context"
+	"net/http"
+
+	"github.com/segmentio/stats"
+)
+
+// RequestWithTags returns a shallow copy of req with its context changed with this provided tags
+// so the they can be used later during the RoundTrip in the metrics recording.
+// The provided ctx must be non-nil.
+func RequestWithTags(req *http.Request, tags ...stats.Tag) *http.Request {
```
f2prateek: Is there a use case where you'd want just the behavior of `RequestWithTags`, but not `AddTagsToRequest`? I'm wondering if we should just make `RequestWithTags` always append new tags, regardless of what tags (if any) are present.

abraithwaite: Should we make a copy of the slice here?  Does it matter since we're not being passed a pointer?  I imagine this slice will be safe to take the address of because it was copied on the stack when calling RequestWithTags, but we might want to make a copy in case the caller changes the underlying slice.

abraithwaite: The problem is that `RequestWithTags` creates a shallow clone, necessary for passing that request into child handlers or transports.  `AddTagsToRequest` explicitly doesn't do that.

I think I would have created the API slightly differently (`RequestWithTags` just an initialization function) to make it less confusing, but I think we should maintain the backwards compatibility.

httpstats/handler.go
```diff
@@ -118,5 +118,5 @@ func (w *responseWriter) complete() {
 	}
 
 	w.metrics.observeResponse(res, "write", w.bytes, now.Sub(w.start))
-	w.eng.ReportAt(w.start, w.metrics)
+	w.eng.ReportAt(w.start, w.metrics, Tags(w.req)...)
```
abraithwaite: This would be more efficient to not take the copy here and reference it directly.  Probably safe enough since this is all in the same package.

context.go
```diff
@@ -0,0 +1,43 @@
+package stats
+
+import "context"
```
abraithwaite: I feel like we need some `sync` in this file now.

httpstats/context_test.go
```diff
@@ -0,0 +1,37 @@
...
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/segmentio/stats"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestContextTags(t *testing.T) {
```
abraithwaite: I want to keep this test since it's exercising the use case I care about.

context.go
```diff
@@ -0,0 +1,43 @@
...
+
+func ContextWithTags(ctx context.Context, tags ...Tag) context.Context {
+	if x := getTagSlice(ctx); x != nil {
+		*x = append(*x, tags...)
+		return ctx
```
achille-roussel: I think we should pass the tags by value and make a new slice and a new context every time here, otherwise the memory that is implicitly shared will be a fertile ground for bugs that are hard to track down.

For example:
```go
return context.WithValue(ctx, contextKeyReqTags, append(ContextTags(ctx), tags...))
```

httpstats/context.go
abraithwaite: After promoting context to the stats package at @achille-roussel 's suggestion, this comes much more naturally.  :raised_hands:

context.go
```diff
@@ -0,0 +1,43 @@
...
+	return nil
+}
+
+// tagsKey is a value for use with context.WithValue. It's used as
+// a pointer so it fits in an interface{} without allocation. This technique
+// for defining context keys was copied from Go 1.7's new use of context in net/http.
+type tagsKey struct{}
+
+// String is Stringer implementation
+func (k *tagsKey) String() string {
```
achille-roussel: We could pass `tagsKey` by value as well, it's an empty struct so it's just a language construct, it doesn't use any memory and its address is a constant.

httpstats/context.go
```diff
@@ -0,0 +1,19 @@
...
+
+// RequestWithTags returns a shallow copy of req with its context changed with
+// this provided tags so the they can be used later.  The provided ctx must be
+// non-nil.  If the context value has already been initialized, it appends the
+// tags to the existing reference and returns the existing Request.
+func RequestWithTags(req *http.Request, tags ...stats.Tag) *http.Request {
+	return req.WithContext(stats.ContextWithTags(req.Context(), tags...))
+}
+
+func Tags(req *http.Request) []stats.Tag {
```
achille-roussel: Should we rename this to `RequestTags` for consistency with the name of `ContextWithTags`/`ContextTags` ?

context.go
abraithwaite: This was copied from the previous implementation, but I had the same thought.

context.go
abraithwaite: I'm on board with creating a new context each time, but the pointer to slice is fundamental to this feature.  Otherwise no child contexts can update the reference and the root httpstats Handler can't get the tags from the child handlers.

abraithwaite: I think once we add some sync access to this key, it should be fine.  We're not exporting anything besides the functions so it will "just work" to the consumers.  If people start changing things in `context.go`, that's when we'll start to have issues.  Otherwise, I think we'll be alright.

Maybe we make a copy of the original slice just to be extra safe that nothing outside of this has a reference to the original slice address.  Callers might have access to the slice contents, but they won't be able to access the slice itself.

achille-roussel: OK, I think this model has different semantics than the way context functions named `With*` usually work.

How about having two functions then?

- `WithContextTags` is the immutable form, always returns a new context
- `AddContextTags` is the mutable form, returns nothing, and always modifies the context passed as argument (panic or do nothing if the context has no tags?)

Splitting the two use cases makes each API harder to use wrong.

context.go
```diff
@@ -0,0 +1,43 @@
...
+		return ret
+	}
+	return nil
+}
+
+func getTagSlice(ctx context.Context) *[]Tag {
+	if tags, ok := ctx.Value(contextKeyReqTags).(*[]Tag); ok {
+		return tags
+	}
+	return nil
```
achille-roussel: This function can be simplified to leverage the zero-value of a pointer type:
```go
tags, _ := ctx.Value(contextKeyReqTags).(*[]Tag)
```
Which also makes its existence questionable.

abraithwaite: Yeah, but this is slightly more clear.  I don't think it needs to be simplified unless you think it'll result in some optimizations or something.

context.go
abraithwaite: Updated the API to match the above ^

abraithwaite:
Thanks!  No rush to merge this so I'll let @systemizer and @Pryz take a look when they get back.

PR #114, opened 2019-11-12
taskstats: use upstream again

### LINKED ISSUE #113: Update segmentio/taskstats dependecy in go.mod
Current version of `github.com/segmentio/taskstats` points to an old commit `v0.0.0-20180727163836-237d1d6b109d` inside the [go.mod](https://github.com/segmentio/stats/blob/master/go.mod#L10) file

However, that commit expects [mdlayher/netlink](https://github.com/mdlayher/netlink) to contain some constants defined that were later renamed.

When compiling my application I had to override the version of `taskstats` in my `go.mod` file with the latest commit  like this

```
github.com/segmentio/taskstats v0.0.0-20190328215536-52f398ff659c
```
I'd suggest doing the same in this `go.mod` file

abraithwaite:
This was fixed in:

https://github.com/mdlayher/taskstats/pull/6

But as a new method TGID instead of inside the PID func.

FYI @jaimemartinez88

Fixes #113


jaimemartinez88:
thanks for the quick fix!

PR #118, opened 2020-04-17
Add zone information in netstats via vpcinfo

Pryz:
We built the `vpcinfo` system to be able to get information about the network where the service is running via DNS. This is a nice way to propagate low level information without having to carry the configuration in the application.

This PR leverage `vpcinfo` in the `netstats` package to add zone information and identify if the traffic is happening cross AZ. This is mostly reusing the work done in Cv3.


Pryz:
Thinking about adding some examples before merging so that it will be easier to share with the teams.

PR #122, opened 2020-07-31
Rename Go version tag from `version` to `go_version`

colinking:
Most of our services will set a version tag that represents the version of their service. However, the Go stats reported by the `procstats` package will overwrite this tag with the current Go runtime version.

This becomes a bigger issue when scraping these stats via Prometheus, since it will consider a sample as invalid if labels are not unique:

image

This PR makes a breaking change to resolve this issue, by renaming the tag that specifics the Go runtime version from `version` to `go_version`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,15 +2,13 @@
 
 A Go package for abstracting stats collection.
 
-Installation
-------------
+## Installation
 
 ```
 go get github.com/segmentio/stats
 ```
 
-Migration to v4
----------------
+## Migration to v4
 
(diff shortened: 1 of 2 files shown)


colinking:
@achille-roussel added a note to the README here: https://github.com/segmentio/stats/tree/colin-go-version#processes

LMK if that looks good to you!

I'll tag as v4.6.0 and add a note on this change in the release notes.

colinking:
Ugh my formatter had some fun with the README... one sec

colinking:
Fixed!

PR #123, opened 2020-09-14
Add Datadog UDS support

sebcante:
### Goal

Adds Support for connecting to datadog via Unix Domain socket  (UDS) instead of UDP

### Why
UDS could be useful :
```
* Bypassing the networking stack brings a significant performance improvement for high traffic.
* While UDP has no error handling, UDS allows the Agent to detect dropped packets and connection errors, while still allowing a non-blocking use.
* DogStatsD can detect the container from which metrics originated and tag those metrics accordingly.
```
Taken from https://docs.datadoghq.com/developers/dogstatsd/unix_socket/?tab=host ^

### Usage to connect to `unixgram` socket instead of `udp`

```
client := NewClient("unixgram:///dir/filename.sock") // unixgram
```
instead of 

```
client := NewClient("localhost:8125") // udp
```

### What

I hope this PR could be useful, completely open for feedbacks to align this PR to the conventions/style of this library. I gave it a first shot.
I inlined comments/questions

### Credit

Datadog [official client lib](https://github.com/DataDog/datadog-go/blob/3255e6186e83fad1e447573c9fa03dd13c023394/statsd/uds.go) , i inspired a lot the code from it.

### DIFF
--- a/datadog/client.go
+++ b/datadog/client.go
@@ -15,7 +15,7 @@ import (
 
 const (
 	// DefaultAddress is the default address to which the datadog client tries
-	// to connect to. By default it connects to UDP
+	// to connect to.
 	DefaultAddress = "localhost:8125"
 
 	// DefaultBufferSize is the default size for batches of metrics sent to
@@ -38,7 +38,7 @@ var (
 // The ClientConfig type is used to configure datadog clients.
 type ClientConfig struct {
 	// Address of the datadog database to send metrics to.
-	// UDP: host:port
+	// UDP: host:port (default)
 	// UDS: unixgram://dir/file.ext
 	Address string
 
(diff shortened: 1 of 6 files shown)


tony-assembly:
is it possible to get someone to review this at all?

sebcante:
We have been using UDS support in production (this PR) for more than  6 months at mx51 it definitely helped us to better manage datadog agent maintenance and go apps reconnecting automatically to it. Something that UDP caused us trouble in the past.
@achille-roussel  any thoughts if this contribution could be merged ?

kevinburkesegment:
Thank you for this! We just merged this upstream and it is part of the v5.1.0 release.

Issue #120, opened 2020-05-28
Support Datadog Distribution metric type ?


sebcante:
Hi @achille-roussel,

Love this `stats` lib. Would you be interested if i attempt to make a PR to support Datadog [Distribution metric types](https://docs.datadoghq.com/developers/metrics/types/?tab=distribution#metric-types) ? 
This allow global percentiles for any tag combination in Datadog.

From this lib point of view it would be exactly as Histogram but the Datadog client would send with a `d` instead of `h` for histogram. 

I am thinking adding a new metric type called `distribution`. 

Let me know thanks!
-seb

achille-roussel:
Hello @sebcante!

We always welcome new contribution, thanks for starting the conversation!

I think there are two ways to go about this:
* since it's a datadog-specific and is similar to histograms, we could make it a configuration option of `datadog.Client`, the upside is it doesn't require API changes in programs, but it prevents using both histograms and distributions on the same datadog client.
* we add a new API to emit distribution metrics, we'll have to think about how to represent those in backends other than datadog.

What do you think?

sebcante:
Thanks @achille-roussel very good pointers 👍 !

It is tempting to add a config in the `datadog.Client` to specify to use datadog distribution for metric type `Histogram` as it is an implementation details of datadog indeed.
That said from what i could understand Distribution comes with limitations:
* we can use only up to 10 tags for a specific metric for percentile aggregation
* we have to [preregister](https://docs.datadoghq.com/metrics/distributions/#aggregations) the aggregation in DD dashboard 

I am also enquiring at the moment with Datadog if it has a cost implication, that might push the consumer of this lib to only hand pick a couple histogram as distribution.
I ll comeback to you shortly when i have the full picture once i talk to Datadog. Stay tuned

bhavanki:
Resolved, at least as a start, via https://github.com/segmentio/stats/pull/132 . Instead of using a simple flag to send all histograms either as histograms or distributions, this change lets you pick prefixes for histogram metric names to be sent as distributions. This way you can have both types.

There are probably other ways to do this, including supporting distributions fully across the stats library, so other / better ideas are definitely welcome.

PR #132, opened 2021-08-11
Support Datadog distribution metric type

bhavanki:
The Datadog client now has the ability to send histogram metrics as the
Datadog-specific distribution metric type. The Datadog client
configuration has a new `DistributionPrefixes` item which specifies the
prefixes of metric names that, when reported as histograms to the stats
library, are to be sent as distributions instead. For example, when the
prefix list is set to `{ "dist_" }`, then any histogram metric whose
name begins with "dist_" is sent as a distribution; all other histograms
are sent as ordinary histograms, as before.

The default configuration sends no histograms as distributions.

### DIFF
--- a/datadog/measure.go
+++ b/datadog/measure.go
@@ -3,20 +3,24 @@ package datadog
 import (
 	"math"
 	"strconv"
+	"strings"
 
 	"github.com/segmentio/stats/v4"
 )
 
+// Datagram format: https://docs.datadoghq.com/developers/dogstatsd/datagram_shell
+
 // AppendMeasure is a formatting routine to append the dogstatsd protocol
 // representation of a measure to a memory buffer.
 func AppendMeasure(b []byte, m stats.Measure) []byte {
-	return AppendMeasureFiltered(b, m, nil)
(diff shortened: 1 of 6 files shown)


bhavanki:
This should address https://github.com/segmentio/stats/issues/120 .

datadog/measure.go
```diff
@@ -3,20 +3,28 @@ package datadog
...
+
 // AppendMeasure is a formatting routine to append the dogstatsd protocol
 // representation of a measure to a memory buffer.
 func AppendMeasure(b []byte, m stats.Measure) []byte {
-	return AppendMeasureFiltered(b, m, nil)
+	return appendMeasureDist(b, m, []string{})
+}
+
+func appendMeasureDist(b []byte, m stats.Measure, distPrefixes []string) []byte {
+	return AppendMeasureFiltered(b, m, nil, distPrefixes)
```
Pryz, line 20: This function looks un-necessary, right ?

bhavanki: You're right, it can be removed. Done!

PR #133, opened 2021-08-12
Fix use of reflect.StringHeader in prometheus/metric.go

bhavanki:
The `go vet` command in Go 1.16 reports a warning for inappropriate
use of reflect.StringHeader.

https://github.com/golang/go/issues/40701

Its use in prometheus/metric.go to convert a byte array to a string in
place began to trigger the warning. That code has been replaced with a
safer variant that avoids the `vet` warning and still converts the array
without allocating new memory.

https://stackoverflow.com/a/66865482

### DIFF
--- a/prometheus/metric.go
+++ b/prometheus/metric.go
@@ -402,10 +402,15 @@ func le(buckets []stats.Value) string {
 		b = appendFloat(b, valueOf(v))
 	}
 
-	return *(*string)(unsafe.Pointer(&reflect.StringHeader{
-		Data: uintptr(unsafe.Pointer(&b[0])),
-		Len:  len(b),
-	}))
+	// This code block converts the byte array to a string without additional
+	// memory allocation.
+	// Source: https://stackoverflow.com/a/66865482 (license: CC BY-SA 4.0)
+	sliceHeader := (*reflect.SliceHeader)(unsafe.Pointer(&b))
+	var s string
+	sh := (*reflect.StringHeader)(unsafe.Pointer(&s))
+	sh.Data = sliceHeader.Data
+	sh.Len = sliceHeader.Len
+	return s
 }
 
 func nextLe(s string) (head string, tail string) {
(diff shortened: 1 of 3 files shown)


prometheus/metric.go
```diff
@@ -401,11 +401,19 @@ func le(buckets []stats.Value) string {
...
+	return unsafeByteSliceToString(b)
+}
 
-	return *(*string)(unsafe.Pointer(&reflect.StringHeader{
-		Data: uintptr(unsafe.Pointer(&b[0])),
-		Len:  len(b),
-	}))
+// This function converts the byte array to a string without additional
+// memory allocation.
+// Source: https://stackoverflow.com/a/66865482 (license: CC BY-SA 4.0)
```
erikdw, line 409: TIL there is a license associated with stack overflow code!  Crazy.
https://stackoverflow.com/help/licensing

Pryz (approved):
LGTM! thanks for adding these tests as well :)

bhavanki:
> LGTM! thanks for adding these tests as well :)

Credit to @collinvandyck for that!

prometheus/metric_test.go
```diff
@@ -11,6 +11,47 @@ import (
 	"github.com/segmentio/stats/v4"
 )
 
+func TestUnsafeByteSliceToString(t *testing.T) {
```
erikdw, line 14: Thanks for adding a test for this!

erikdw (approved):

prometheus/metric.go
bhavanki: It caused quite the kerfuffle a few years ago.

prometheus/metric_test.go
bhavanki: This is @collinvandyck 's contribution, kudos to him!

erikdw: Hehe, yeah, after submitting my review I saw the same comment from Julien :P

PR #136, opened 2022-02-02
emit metrics for deadline_exceeded

dk1027:
When Go HTTP Server's timeouts are set and are exceeded, read/write will return the `os.ErrDeadlineExceeded` error and this causes the ALB to return a 502. This PR tags `conn.error.count` metrics with `error=deadline_exceeded` when the error is `os.ErrDeadlineExceeded`. This allows better visibility when ALB shows 502 errors.

### DIFF
--- a/netstats/conn.go
+++ b/netstats/conn.go
@@ -4,6 +4,7 @@ import (
 	"io"
 	"math"
 	"net"
+	"os"
 	"strconv"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


bhavanki (commented):
(I'm setting a higher bar here than usual because this is a publicly available library.

The change makes sense, but I can see that an "error" tag isn't used anywhere else in this library, so it would be a somewhat significant addition to the public contract. It seems compatible, but could users already be adding on their own "error" tags that will interfere here? Are there other ways to make the information available?

I can see that listener.go has similar error-handling code. Should this change be done there too?

Lastly, can this get unit test coverage?

Pryz:
Is this change still needed ?

PR #137, opened 2022-04-29
Add an OpenTelemetry handler

Pryz:
This PR adds a first version of an OpenTelemetry handler. This handler
implement the otlp protocol with an HTTP Client.

There are few caveats with the current implementation:
- It only supports HTTP but GRPC should be easy enough to add later if needed.
- The client is written in a fire-and-forget fashion, meaning that metrics
will be sent once even if there is a failure when sending the request.
- The handler only support backends which expect cumulative values such as Prometheus.


otlp/client.go
```diff
@@ -0,0 +1,91 @@
...
+func NewHTTPClient(endpoint string) *HTTPClient {
+	return &HTTPClient{
+		//TODO: add sane default timeout configuration.
+		client:   http.DefaultClient,
+		endpoint: endpoint,
+	}
+}
+
+func (c *HTTPClient) Handle(ctx context.Context, request *colmetricpb.ExportMetricsServiceRequest) error {
+	println(request)
```
kevinburkesegment, line 38: Stray?

otlp/client.go
```diff
@@ -0,0 +1,91 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to marshal request: %s", err)
+	}
+
+	httpReq, err := newRequest(c.endpoint, rawReq)
+	if err != nil {
+		return fmt.Errorf("failed to create HTTP request: %s", err)
+	}
+
+	return c.do(httpReq)
```
kevinburkesegment, line 49: Shouldn't we pass through the ctx.Context here? or do httpReq = httpReq.WithContext(ctx) ? 

Otherwise what's the point of passing it through?

otlp/client.go
```diff
@@ -0,0 +1,91 @@
...
+	return nil
+}
+
+func newRequest(endpoint string, data []byte) (*http.Request, error) {
+	u, err := url.Parse(endpoint)
+	if err != nil {
+		return nil, err
+	}
+
+	req, err := http.NewRequest(http.MethodPost, u.String(), nil)
```
kevinburkesegment, line 82: NewRequestWithContext uses better defaults here IIRC.

otlp/client.go
```diff
@@ -0,0 +1,91 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	req.Header.Set("Content-Type", "application/x-protobuf")
```
kevinburkesegment, line 87: User-agent?

otlp/handler.go
```diff
@@ -0,0 +1,188 @@
...
+)
+
+const (
+	// DefaultMaxMetrics is the default maximum of metrics kept in memory
+	// by the handler.
+	DefaultMaxMetrics = 5000
+
+	// DefaultFlushInterval is the default interval to flush the metrics
+	// to the OpenTelemetry destination.
+	DefaultFlushInterval = 10 * time.Second
```
kevinburkesegment, line 23: How do these interact? Will we flush faster, or start dropping messages, if we get to 5000 before ten seconds are up

otlp/handler.go
```diff
@@ -0,0 +1,188 @@
...
+)
+
+// Handler implement stats.Handler to be used to forward metrics to an
```
kevinburkesegment, line 26:
```suggestion
// Handler implements stats.Handler to be used to forward metrics to an
```

otlp/handler.go
```diff
@@ -0,0 +1,188 @@
...
+// OpenTelemetry destination. Usually an OpenTelemetry Collector.
+//
+// With the current implementation this Handler is targeting a Prometheus
+// based backend or any backend expecting cumulative value.
```
kevinburkesegment, line 30:
```suggestion
// based backend or any backend expecting cumulative values.
```

otlp/handler.go
```diff
@@ -0,0 +1,188 @@
...
+		}
+		metrics = append(metrics, convertMetrics(*m)...)
+		m.flushed = true
+	}
+
+	if len(metrics) == 0 {
+		return nil
+	}
+
+	//FIXME how big a metrics service requests can be ? need pagination ?
```
kevinburkesegment, line 147:
```suggestion
	//FIXME how big can a metrics service request be ? need pagination ?
```

otlp/measure.go
```diff
@@ -0,0 +1,122 @@
...
+)
+
+func convertMetrics(metrics ...metric) []*metricpb.Metric {
+	mm := []*metricpb.Metric{}
+
+	for _, metric := range metrics {
+		attributes := tagsToAttributes(metric.tags...)
+
+		m := &metricpb.Metric{
+			Name: fmt.Sprintf("%s.%s", metric.measureName, metric.fieldName),
```
kevinburkesegment, line 18: If it's just strings i'd just do `metric.measureName + "." + metric.FieldName` and avoid the Sprint parse overhead.

kevinburkesegment (approved):

otlp/client.go
Pryz: thanks

otlp/client.go
Pryz: yep forgot to pass it through :) thanks

Pryz:
Thanks for the review @kevinburkesegment !

otlp/handler.go
Pryz: Updated the doc to explain the behavior :)

otlp/client.go
Pryz: good idea. done

otlp/handler.go
```diff
@@ -0,0 +1,212 @@
...
+const (
+	// DefaultMaxMetrics is the default maximum of metrics kept in memory
+	// by the handler.
+	DefaultMaxMetrics = 5000
+
+	// DefaultFlushInterval is the default interval to flush the metrics
+	// to the OpenTelemetry destination.
+	//
+	// Metrics will be flushed to the destination when DefaultFlushInterval or
+	// DefaultMaxMetrics are reach. Whichever comes first.
```
kevinburkesegment, line 26:
```suggestion
	// DefaultMaxMetrics are reached, whichever comes first.
```

otlp/handler.go
```diff
@@ -0,0 +1,212 @@
...
+	DefaultFlushInterval = 10 * time.Second
+)
+
+// Handler implements stats.Handler to be used to forward metrics to an
+// OpenTelemetry destination. Usually an OpenTelemetry Collector.
+//
+// With the current implementation this Handler is targeting a Prometheus
+// based backend or any backend expecting cumulative values.
+//
+// This Handler leverages a double linked list with a map to implement
```
kevinburkesegment, line 36:
```suggestion
// This Handler leverages a doubly linked list with a map to implement
```

PR #138, opened 2022-04-29
Add github workflow

Pryz:


influxdb/client_test.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/segmentio/stats/v4"
 )
 
-func TestClient(t *testing.T) {
```
Pryz, line 14: Not sure if we want to keep this. I don't think the code has been exercised in a while.

kevinburkesegment (approved):
Nice! Is it worth adding go vet ./... as well?

Pryz:
I've added golangci to add more tests. Running it against master only for now but once we've fixed most of the failures we can enable it against all PRs.

PR #139, opened 2022-09-14
TEO11Y-60 Support For Datadog Distributions

boggsboggs:
This PR makes two semantic changes:
- Adds a FilteredHandler to the stats package that processes measures with a specified filter function before forwarding to another handler
- Adds the UseDistributions flag to datadog.ClientConfig which sends all histogram measures as distributions

I did some refactoring in the datadog package to facilitate this change:
- Converted `datadog.AppendMeasureFiltered` into a method on `datadog.serializer` named `AppendMeasure`
- Moved `datadog.serializer` from `datadog/client.go` into `datadog/measures.go`
- Renamed `datadog/measures.go` to `datadog/serializer.go`

Because of the mix of refactoring and code additions, I expect this PR will be easiest to review one commit at a time.


datadog/client.go
```diff
@@ -52,6 +50,9 @@ type ClientConfig struct {
 	// Set of name prefixes for metrics to be sent as distributions instead of
 	// as histograms.
 	DistributionPrefixes []string
+
+	// Flag indicating whether to send histograms as `d` type or `h` type
```
kevinburkesegment, line 54: This doc doesn't make clear whether "true" will send the "d" type or the "h" type. It could also be clearer what the "d" type and the "h" type are - maybe link to a doc?

datadog/client_test.go
```diff
@@ -58,6 +60,47 @@ func TestClientWithDistributionPrefixes(t *testing.T) {
 	}
 }
 
+func TestClientWithUseDistributions(t *testing.T) {
+
+	// Start a go routine listening for packets and giving them back on packets chan
```
kevinburkesegment, line 65:
```suggestion
	// Start a goroutine listening for packets and giving them back on packets chan
```

datadog/client_test.go
```diff
@@ -125,3 +168,29 @@ func BenchmarkClient(b *testing.B) {
 		})
 	}
 }
+
+// startUDPListener starts a go routine listening for UDP packets on 127.0.0.1 and an available port.
```
kevinburkesegment, line 172:
```suggestion
// startUDPListener starts a goroutine listening for UDP packets on 127.0.0.1 and an available port.
```

datadog/client_test.go
```diff
@@ -125,3 +168,29 @@ func BenchmarkClient(b *testing.B) {
...
+// The address listened to is returned as `addr`. The payloads of packets received are copied to `packets`
+func startUDPListener(t *testing.T, packets chan []byte) (addr string, closer io.Closer) {
```
kevinburkesegment, line 174: add `t.Helper()` here?

handler.go
```diff
@@ -74,6 +74,30 @@ func (m *multiHandler) Flush() {
 	}
 }
 
+// FilteredHandler constructs a Handler that processes Measures with `filter` before forwarding to `h`
+func FilteredHandler(h Handler, filter func([]Measure) []Measure) Handler {
```
kevinburkesegment, line 78: Another API here would be `func(m Measure) bool`, but maybe you can't do that because you need the ability to compare multiple Measure ments. 

I'm nervous that this API might involve a lot of copying/allocations.

handler.go
```diff
@@ -74,6 +74,30 @@ func (m *multiHandler) Flush() {
...
+
+type filteredHandler struct {
+	handler Handler
+	filter  func([]Measure) []Measure
+}
+
+func (h *filteredHandler) HandleMeasures(time time.Time, measures ...Measure) {
+	filteredMeasures := h.filter(measures)
+
+	if len(filteredMeasures) == 0 {
```
kevinburkesegment, line 90: This doesn't seem obviously correct... it's possible there are downstream handlers that might want to get called even if all measurements have been filtered out.

datadog/serializer.go
```diff
@@ -0,0 +1,180 @@
...
+		var splitIndex int
+
+		for splitIndex != len(b) {
+			i := bytes.IndexByte(b[splitIndex:], '\n')
+			if i < 0 {
+				panic("stats/datadog: metrics are not formatted for the dogstatsd protocol")
+			}
+			if (i + splitIndex) >= s.bufferSize {
+				if splitIndex == 0 {
+					log.Printf("stats/datadog: metric of length %d B doesn't fit in the socket buffer of size %d B: %s", i+1, s.bufferSize, string(b))
```
kevinburkesegment, line 49: Maybe explain in the log message what we actually do in this case? (Truncate the message it looks like)

datadog/serializer.go
```diff
@@ -0,0 +1,180 @@
...
+		}
+
+		n += c
+		b = b[splitIndex:]
+	}
+
+	return n, nil
+}
+
+func (s *serializer) close() {
```
kevinburkesegment, line 70: No error in the signature here? Seems better to return it and let the caller decide to ignore it.

datadog/serializer.go
```diff
@@ -0,0 +1,180 @@
...
+		b = s.AppendMeasure(b, m)
+	}
+	return b
+}
+
+// AppendMeasure is a formatting routine to append the dogstatsd protocol
+// representation of a measure to a memory buffer.
+// Tags listed in the s.filters are removed. (some tags may not be suitable for submission to DataDog)
+// Histogram metrics will be sent as distribution type if the metric name matches s.distPrefixes
+func (s *serializer) AppendMeasure(b []byte, m stats.Measure) []byte {
```
kevinburkesegment, line 87: Might be nice to link the doc reference here as well.

datadog/serializer.go
boggsboggs: These lines are green only because I moved them. I didn't actually write this code, and I'd like to keep this change as simple as possible by not editing the existing implementation.

datadog/serializer.go
boggsboggs: Ditto

These lines are green only because I moved them. I didn't actually write this code, and I'd like to keep this change as simple as possible by not editing the existing implementation.

datadog/serializer.go
```diff
@@ -0,0 +1,180 @@
...
+	for _, m := range measures {
+		b = s.AppendMeasure(b, m)
+	}
+	return b
+}
+
+// AppendMeasure is a formatting routine to append the dogstatsd protocol
+// representation of a measure to a memory buffer.
+// Tags listed in the s.filters are removed. (some tags may not be suitable for submission to DataDog)
+// Histogram metrics will be sent as distribution type if the metric name matches s.distPrefixes
```
boggsboggs, line 86:
```suggestion
// Histogram metrics will be sent as distribution type if the metric name matches s.distPrefixes
// DogStatsd Protocol Docs: https://docs.datadoghq.com/developers/dogstatsd/datagram_shell?tab=metrics#the-dogstatsd-protocol
```

datadog/serializer.go
boggsboggs: Will do!

datadog/client_test.go
boggsboggs: I'm not familiar with `t.Helper()` In the docs it says:
"Helper marks the calling function as a test helper function. When printing file and line information, that function will be skipped."

I don't see any other documented effects of calling t.Helper(). Not printing the full stack trace on errors is not behavior I would want. Am I missing something?

handler.go
boggsboggs: Yeah, I agree. It's simpler and probably less prone to future problems to remove that short-circuit. Good suggestion!

handler.go
boggsboggs: Ok, so I imagine a for loop inside of `filter` that allocates a new slice (on the stack) and builds it from the input slice:
```
result []Measure = {}
range _, measure := measures {
  if CONDITION {
    result = append(measure, result)
  }
}
return result
```

I can move that loop up into HandleMeasures, which would look like:
```
filteredMeasures []Measure = {}
range _, measure := measures {
  if filter(measure) {
    filteredMeasures = append(filteredMeasures, result)
  }
}
return h.handler.HandleMeasures(time, filteredMeasures...)
```

It's almost the same code... just a slight change on where the function boundaries are. Given that the code is working as is, I'd rather leave it. Were you thinking of a different implementation? Is there a way to optimize away that stack slice allocation and copy that I don't see?

datadog/client.go
boggsboggs, line 54:
```suggestion
	// UseDistributions True indicates to send histograms with `d` type instead of `h` type
	// https://docs.datadoghq.com/developers/dogstatsd/datagram_shell?tab=metrics#the-dogstatsd-protocol
```

boggsboggs: yeah, this could be more clear. Updated! Thanks!

datadog/client_test.go
kevinburkesegment: You don't get a full stacktrace on t.Fatal anyway, just the failing line. With t.Helper It'll print the line from the test that called it instead of the line from the helper, which I find is useful.

You should be able to verify this by manually making the test fail and then running `go test ./...`

handler.go
kevinburkesegment: No that's fine, you're right, this is probably thinking too hard

kevinburkesegment (approved):

datadog/client_test.go
boggsboggs: Oh, I see how that's useful now. Already merged the PR, but I know why you'd use t.Helper for the future now. Thanks!

PR #141, opened 2022-11-15
Create dedicated go module for otlp exporter

Pryz:


achille-roussel (approved):

boggsboggs:
Checked that this change does, in fact, allow me to build orchestration-worker with this sha as a dependency.  Thanks for unblocking me @Pryz!

PR #143, opened 2023-01-20
Fix deadlock scenario

vertoforce:


vertoforce:
Apologies, accidentally opened this PR

PR #147, opened 2023-07-15
tag sort dedup

extemporalgenome:
This is meant to be read commit-by-commit.

- tag.go: conventional test naming.
- add golang.org/x/exp/slices dependency
- tag.go: use slices package Sort functionality
- SortTags deduplicates, uses slices.SortStable


tag.go
```diff
@@ -47,21 +47,54 @@ func SortTags(tags []Tag) []Tag {
...
+		// No input tags had non-empty names:
+		// return nil to be consistent for ease of testing.
+		return nil
+	}
 
-func concatTags(t1 []Tag, t2 []Tag) []Tag {
+	return out
+}
+
+// mergeTags returns the sorted, deduplicated-by-name union of t1 and t2.
```
kevinburkesegment, line 86: Can you explain in the comment what strategy we use to deduplicate (last write wins).

.github/workflows/go.yml
```diff
@@ -1,30 +1,37 @@
...
 
-    - name: Download Dependencies
-      run: go mod download
+      - name: Identify OS
+        run: uname -a
 
-    - name: Run Tests
-      run: go test -race -tags=${{ matrix.tags }} ./...
+      - name: Identify Go Version
+        run: go version
```
kevinburkesegment, line 31: These start a new container per command so there's a fair amount of overhead. How about e.g.

```yml
name: Identify Host 
run: |
    uname -a
    go version
```

.github/workflows/golangci-lint.yml
```diff
@@ -15,13 +16,13 @@ jobs:
...
-          go-version: ^1.18
+          go-version: '>=1.20'
 
       - uses: actions/checkout@v3
 
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v3.1.0
         with:
-          version: v1.45.2
+          version: v1.53
```
kevinburkesegment, line 28: Thanks for updating this.

kevinburkesegment (approved):

.github/workflows/go.yml
extemporalgenome: For BuildKite that'd definitely be the case, but for Github Actions, it looks like jobs create their own container/VM/whatever, but steps are just separate OS processes in that same container.

As such, steps look like they're no more expensive than running separate commands in a single shell. By making them separate steps, they do create nice UI sections though, and the above info was useful for debugging the especially stale workflow config.

tag.go
extemporalgenome: The latest commit has hopefully addressed this. Please re-review.

kevinburkesegment (approved):

PR #149, opened 2023-07-18
lint fixes

extemporalgenome:
- .golangci.yml: disable broken linter
- golangci-lint run --fix
- hand-fix some linting issues
- just use errgroup instead of reinventing an errgroup


context.go
```diff
@@ -62,12 +62,12 @@ type tagSlice struct {
 // for defining context keys was copied from Go 1.7's new use of context in net/http.
 type tagsKey struct{}
 
-// String is Stringer implementation
+// String is Stringer implementation.
```
kevinburkesegment, line 65:
```suggestion
// String implements the fmt.Stringer interface.
```

kevinburkesegment (approved):

PR #151, opened 2023-08-14
syscall.Sysinfo returns an uint32 on linux/arm, but uint64 otherwise

noctarius:
Seems like the syscall result isn't the same between arm (32bit) and arm64. Totalram can be an uint32. Instead of creating an additional build constraint, I just added the cast which will be removed when unnecessary, but stays in place when required.


extemporalgenome:
Thanks!

PR #152, opened 2023-08-29
fix: update the x/exp package, and updated tagIsLess function

needcaffeine:
slices.SortStableFunc now requires the cmp function to return an int, not bool.


tag.go
```diff
@@ -51,7 +51,15 @@ func SortTags(tags []Tag) []Tag {
 	return deduplicateTags(tags)
 }
 
-func tagIsLess(a, b Tag) bool { return a.Name < b.Name }
+func tagIsLess(a, b Tag) int {
```
extemporalgenome, line 54: I recommend you call this `compareTags`, since `less` implies a returned bool, which is no longer the case.

extemporalgenome (approved):

PR #156, opened 2023-12-06
tag: fix build

kevinburkesegment:
On tip (the Go 1.22 code freeze), Go reports that `compareTag` has the wrong method signature:

```
type func(a Tag, b Tag) int of tagCompare does not match inferred type func(a Tag, b Tag) bool for func(a E, b E) bool
```

Fix this by using the correct method signature.


kevinburkesegment:
turns out we needed to update the x/exp dependency, which has been done.

PR #159, opened 2024-06-19
engine: report go version on startup

kevinburkesegment:
The first time you report any metric, additionally report the running Go version. Add two flags that can be used to disable this behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -385,3 +385,16 @@ func main() {
     // ...
 }
 ```
+
+### Addendum
+
+By default, the stats library will report the running go version when you
+invoke NewEngine() as three metrics:
+
+- `go_version.major`
+- `go_version.minor`
+- `go_version.patch`
+
+Set `STATS_DISABLE_GO_VERSION_REPORTING` to `true` in your environment, or set
+`stats.GoVersionReportingEnabled` to `false` before collecting any metrics, to
+disable this behavior.
--- a/README.md
+++ b/README.md
@@ -183,7 +183,7 @@ Monitoring
 
 > 🚧 Go metrics reported with the `procstats` package were previously tagged with a
 > `version` label that reported the Go runtime version. This label was renamed to
-> `go_version` in v5.6.0.
+> `go_version` in v4.6.0.
 
 The
 [github.com/segmentio/stats/procstats](https://godoc.org/github.com/segmentio/stats/procstats)
@@ -389,11 +389,11 @@ func main() {
 ### Addendum
 
 By default, the stats library will report the running go version when you
-invoke NewEngine() as three metrics:
+invoke NewEngine() as a metric:
 
-- `go_version.major`
-- `go_version.minor`
-- `go_version.patch`
+- `go_version` with value 1 and a `tag` set to the current version.
+- `stats_version` with value ` and a `tag` set to the tag value of
+  segmentio/stats.
 
 Set `STATS_DISABLE_GO_VERSION_REPORTING` to `true` in your environment, or set
 `stats.GoVersionReportingEnabled` to `false` before collecting any metrics, to
--- a/engine.go
+++ b/engine.go
@@ -1,19 +1,22 @@
 package stats
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 	"reflect"
+	"runtime"
+	"strings"
 	"sync"
 	"time"
 )
 
-// An Engine carries the context for producing metrics, it is configured by
+// An Engine carries the context for producing metrics. It is configured by
 // setting the exported fields or using the helper methods to create sub-engines
 // that inherit the configuration of the base they were created from.
 //
 // The program must not modify the engine's handler, prefix, or tags after it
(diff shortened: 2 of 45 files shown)


README.md
```diff
@@ -385,3 +385,16 @@ func main() {
     // ...
 }
 ```
+
+### Errata
```
mckern, line 389: This would technically be addendum, not errata; it's not correcting anything, just adding.

engine.go
```diff
@@ -138,7 +144,35 @@ func (eng *Engine) ClockAt(name string, start time.Time, tags ...Tag) *Clock {
...
+				if err == nil {
+					eng.measure_(t, "go_version.minor", min, Gauge)
+				}
+				if len(parts) == 3 {
+					patch, err := strconv.Atoi(parts[2])
+					if err == nil {
+						eng.measure_(t, "go_version.patch", patch, Gauge)
+					}
+				}
+			}
```
mckern, line 170: Would it be sufficient to just ship the version string (but keep that trimmed prefix, good looking out) and let folks on the consumer side manage/mangle/parse the value as they see fit? I'm not sure Major or Minor have much utility in isolation.

kevinburkesegment: I didn't see a way to report a string instead of a number, but I will look again

mckern: Yeah fair, I don't know the stats stack intimately so maybe it's just not a thing. Thanks for checking.

README.md
kevinburkesegment: Fixed

ABHINAV-SUREKA (approved):

mckern (commented):
Thank you for working on this. I've gone over this a bit, but I'm kind of torn on this implementation because while code that exists is generally preferable to code that doesn't, I'm still not convinced that this should be shipped as 3 separate values using gauges.

[What you turned up from the procstats subpackage](https://github.com/segmentio/stats/blob/main/procstats/go.go#L114-L118) may be what we really need for our use-cases.

engine.go
etiennep: What if we put the version string in a tag, and set a gauge/counter to `1`? Then you can sum by version tag.

kevinburkesegment:
@etiennep @mckern I reworked this to just publish a single metric with a value of "1" and the go version as a tag.

ABHINAV-SUREKA (dismissed):

engine.go
```diff
@@ -144,7 +149,24 @@ func (eng *Engine) ClockAt(name string, start time.Time, tags ...Tag) *Clock {
 	}
 }
 
+var GoVersionReportingEnabled = os.Getenv("STATS_DISABLE_GO_VERSION_REPORTING") != "true"
```
mckern, line 152: Suggestion: setting boolean environment variables to "yes", "1", etc. is pretty common. Checking for emptiness/"false"/"f"/"0" may be safer than only looking for "true".

engine.go
```diff
@@ -144,7 +149,24 @@ func (eng *Engine) ClockAt(name string, start time.Time, tags ...Tag) *Clock {
...
+		eng.once.Do(func() {
+			vsn := strings.TrimPrefix(runtime.Version(), "go")
+			parts := strings.Split(vsn, ".")
+			// this filters out weird compiled Go versions like tip.
+			// older Go version might be "go1.13"
+			if len(parts) == 2 || len(parts) == 3 {
+				eng.measureOne(t, "go_version", 1, Gauge, []Tag{{"go_version", vsn}}...)
+			}
+		})
+	}
```
mckern, line 165:
```suggestion
			vsn := strings.TrimPrefix(runtime.Version(), "go")
			eng.measureOne(t, "go_version", 1, Gauge, []Tag{{"go_version", vsn}}...)
		})
	}
```

netstats/listener_test.go
```diff
@@ -58,15 +63,31 @@ func TestListenerError(t *testing.T) {
...
-			Tags:   []stats.Tag{stats.T("operation", "accept"), stats.T("protocol", "tcp")},
-		},
+	vsn := strings.TrimPrefix(runtime.Version(), "go")
+	parts := strings.Split(vsn, ".")
+	measures := h.Measures()
+	if len(parts) == 2 || len(parts) == 3 {
+		if len(measures) != 1+1 {
+			t.Fatalf("expecting to get %d metrics, got back %d: %v", 1+1, len(measures), measures)
+		}
+	}
```
mckern, line 73: I don't think this tests the new implementation?

procstats/go.go
```diff
@@ -109,7 +109,7 @@ func NewGoMetrics() *GoMetrics {
 	return NewGoMetricsWith(stats.DefaultEngine)
 }
 
-// NewGoMetricsWith creates a new collector for the Go unrtime that producers
+// NewGoMetricsWith creates a new collector for the Go runtime that producers
```
mckern, line 112: Big-Lebowski

etiennep (dismissed):

etiennep (approved):

PR #175, opened 2024-12-18
debugstats: new package

kevinburkesegment:
This makes it easier to troubleshoot problems with the metrics pipeline, by visualizing what metrics are emitted.

Fix the existing go_version and stats_version reporter to ensure that they work even if metrics are reported with ReportAt instead of Incr, Set, Observe etc.


ABHINAV-SUREKA (approved):

etiennep (approved):
LGTM

PR #176, opened 2024-12-19
procstats: bump taskstats and fix file handle leak

mdlayher:
The taskstats file handle is not being closed and thus it is possible for file descriptors to pile up forever. Pull in a much newer version and fix this problem.

In addition, I don't think the naked returns or check helper are doing any favors for readability and have removed them in this block in favor of switch error handling.


kevinburkesegment:
Eek. Thank you. I'm about to go to bed but will review this tomorrow.

kevinburkesegment (approved):

mdlayher:
Thanks Kevin!

kevinburkesegment:
Thanks for the PR!

PR #185, opened 2025-03-10
add buildkite test run

extemporalgenome:


kevinburkesegment:
Great! So people who are not Twilio/Segment employees won't be able to access the Buildkite build but since we've historically not had a lot of contributions to this repo from outside users and the test runner is fairly simple that's not a big problem for me.

Should we delete the Github runner? I don't see the point in having two different test runners do the same thing.

extemporalgenome:
Oh. I didn't realize this is public. @kevinburkesegment if we tend to have public projects running via github actions, and plan to continue that support, then we should close this PR.

extemporalgenome:
I've gone ahead and deleted the buildkite pipeline/integration, and am closing this PR.

PR #189, opened 2025-03-25
PLATOBS-4923 Add logic to replace invalid unicode with '\ufffd'

ABHINAV-SUREKA:
https://twilio-engineering.atlassian.net/browse/PLATOBS-4923

### DIFF
--- a/datadog/serializer.go
+++ b/datadog/serializer.go
@@ -27,6 +27,8 @@ func (s *serializer) Write(b []byte) (int, error) {
 		return 0, io.ErrClosedPipe
 	}
 
+	// Ensure the serialized metric payload have valid UTF-8 encoded bytes
+	b = bytes.ToValidUTF8(b, []byte("\uFFFD"))
 	if len(b) <= s.bufferSize {
 		return s.conn.Write(b)
 	}
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -1,3 +1,4 @@
+ABHINAV SUREKA <7731608+ABHINAV-SUREKA@users.noreply.github.com>
 Achille Roussel <achille.roussel@segment.com>
 Alan Braithwaite <alan@segment.com>
 Amir Abu Shareb <yields@icloud.com>
--- a/datadog/serializer_test.go
+++ b/datadog/serializer_test.go
@@ -1,8 +1,12 @@
 package datadog
 
 import (
+	"bytes"
+	"errors"
+	"strings"
 	"testing"
 	"time"
+	"unicode/utf8"
 
 	stats "github.com/segmentio/stats/v5"
 )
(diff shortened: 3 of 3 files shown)


datadog/utf.go
```diff
@@ -0,0 +1,25 @@
...
+			// Invalid UTF-8 sequence found, append replacement character
+			validBytes = append(validBytes, []byte("\uFFFD")...)
+		} else {
+			// Valid UTF-8 sequence found, append it to the validBytes
+			validBytes = append(validBytes, b[:size]...)
+		}
+		b = b[size:]
+	}
+	return validBytes
+}
```
kevinburkesegment, line 25: Instead of appending one byte at a time how about:

```go
func ensureValidUTF8(b []byte) []byte {
	// Fast path: if already valid, return immediately
	if utf8.Valid(b) {
		return b
	}
	
	// Pre-calculate the replacement bytes once
	replacement := []byte("\uFFFD")
	
	// Initialize with a reasonable capacity
	validBytes := make([]byte, 0, len(b))
	
	// Track the start of the current valid segment
	start := 0
	
	for i := 0; i < len(b); {
		if b[i] < utf8.RuneSelf {
			// Fast path for ASCII (will be valid UTF-8)
			i++
			continue
		}
		
		r, size := utf8.DecodeRune(b[i:])
		if r == utf8.RuneError {
			// Found an invalid sequence, copy the valid segment we've accumulated
			validBytes = append(validBytes, b[start:i]...)
			// Add the replacement character
			validBytes = append(validBytes, replacement...)
			// Skip the invalid byte
			i++
			// Reset the start of the next valid segment
			start = i
		} else {
			// Move past this valid rune
			i += size
		}
	}
	
	// Append any remaining valid segment
	if start < len(b) {
		validBytes = append(validBytes, b[start:]...)
	}
	
	return validBytes
}
```

There are other approaches - search through github.com/golang/go for any code that references utf8.RuneError, for example you could look at encoding/json

datadog/utf_test.go
```diff
@@ -0,0 +1,43 @@
...
+	},
+	// Metric with invalid UTF-8 embedded
+	{
+		input:    []byte("metric_name:\xe2\x26|h|#tag1:\xffvalue;tag2:value2"),
+		expected: []byte("metric_name:\uFFFD&|h|#tag1:\uFFFDvalue;tag2:value2"),
+	},
+	// Metric with invalid UTF-8 at the end
+	{
+		input:    []byte("metric.value:123|g|#\xff\xfe"),
+		expected: []byte("metric.value:123|g|#\uFFFD\uFFFD"),
```
kevinburkesegment, line 30: Let's add a test that valid multi-byte UTF-8 characters - for example, emojis - get passed through correctly.

datadog/serializer.go
```diff
@@ -27,6 +27,8 @@ func (s *serializer) Write(b []byte) (int, error) {
 		return 0, io.ErrClosedPipe
 	}
 
+	b = ensureValidUTF8(b)
```
kevinburkesegment, line 30: Scratch that - we should be using `bytes.ToValidUTF8` here.

datadog/utf_test.go
ABHINAV-SUREKA: sure. will do it.

datadog/utf.go
ABHINAV-SUREKA: Ok. Yeah I was thinking the same. Lemme look into it.

datadog/serializer.go
ABHINAV-SUREKA: this is amazing. didn't know this existed already.

kevinburkesegment (dismissed):

kevinburkesegment:
Can you run `make authors` or `make AUTHORS.txt` to add yourself to the list of authors?

ABHINAV-SUREKA:
> Can you run `make authors` or `make AUTHORS.txt` to add yourself to the list of authors?

Done

kevinburkesegment (approved):

PR #192, opened 2025-05-24
datadog: sanitize metric names, fields, and tags 

kevinburkesegment:


kevinburkesegment:
Benchmarks look a lot better when I remove the unicode logic, about 15% worse than main vs. 300% worse

frajuqeb (approved):

PR #201, opened 2025-06-11
upgrade v4 dep on v5

extemporalgenome:
- **all: reimplement stats/v4 in terms of stats/v5**
- **v4: bump internal dependency on v5 to latest**


go.mod
```diff
@@ -1,29 +1,31 @@
 module github.com/segmentio/stats/v4
 
-go 1.22.0
+go 1.23.0
+
+toolchain go1.24.4
```
kevinburkesegment, line 5: Can we remove this line? This forces a redownload of the entire Go binary for anyone who is running this.

extemporalgenome: Definitely.

extemporalgenome: Fixed in af8303c

kevinburkesegment (approved):

PR #225, opened 2026-07-29
all: fold the otlp package back into the main module

kevinburkesegment:
The nested otlp/go.mod meant `go test ./...`, `go build ./...`, and golangci-lint at the repo root all silently skipped the otlp package, and every dependency bump had to be applied in two places. Remove it and let otlp be an ordinary package of github.com/segmentio/stats/v5.

The cost is the module's minimum Go version: the OpenTelemetry SDK and gRPC both declare `go 1.25.0`, so the root go.mod moves from 1.24.0 to 1.25.0 and every consumer of stats now pulls in the OTel SDK and gRPC. CI's oldstable is already 1.25, so the test matrix is unaffected.

Now that the package is actually linted, fix the 17 issues it had been hiding: missing periods on declaration comments (godot), an import group that ignored the local-prefixes setting (goimports), "signalled" -> "signaled" (misspell), unused parameters (revive), and an unused resourceAttrs field on SDKHandler. recordMetric's metricName parameter was never read, so it is dropped rather than renamed.


wavetylor (approved):

wavetylor:
thank you!

